PR #1, opened 2018-01-25
Remove splashscreen

vitorgrs:
This should remove Splashscreen on Windows 10 16299+. 👍


felixse:
Good idea, thank you for your contribution 😃

PR #41, opened 2018-06-15
Add WSL as default profile & terminal colors used in Ubuntu Bash

Linard11:


skipmeister123:
Have you tested with multiple distros installed?

--
Hans Kokx
On 6/14/2018 8:25:56 PM, Fuzzel <notifications@github.com> wrote:
You can view, comment on, or merge this pull request online at:
  https://github.com/felixse/FluentTerminal/pull/41 [https://github.com/felixse/FluentTerminal/pull/41]
Commit Summary
* Add WSL as default profile
* Add 'Ubuntu Bash' as default theme
File Changes
* M FluentTerminal.App/Services/Implementation/DefaultValueProvider.cs [https://github.com/felixse/FluentTerminal/pull/41/files#diff-0] (43)
* M README.md [https://github.com/felixse/FluentTerminal/pull/41/files#diff-1] (2)
Patch Links:
* https://github.com/felixse/FluentTerminal/pull/41.patch [https://github.com/felixse/FluentTerminal/pull/41.patch]
* https://github.com/felixse/FluentTerminal/pull/41.diff [https://github.com/felixse/FluentTerminal/pull/41.diff]

Linard11:
Yes, but as it will launch `C:\windows\system32\wsl.exe`, whatever linux subsystem is set to default, via `wslconfig.exe`, will be launched. There would also be the possibility to launch specificly e.g. the ubuntu subsystem, but that would have to execute `"%LOCALAPPDATA%\Microsoft\WindowsApps\<subsystem>.exe`, which does work in general, but there were two issues I had with it:

-  Adding it with the open dialogue fails due to some missing access rights giving a warning. But entering the path as a string into the text field does work.
- For the context menu integration, the positional arguments are not working properly, so it will always open the default directory, being the home directory of the user, instead of where the shortcut menu was called. This only seems to work when executing `wsl.exe`, and I'm not proficient enough with what arguments to add in the registry entries of the shell command to make it work.

FluentTerminal.App/Services/Implementation/DefaultValueProvider.cs
```diff
@@ -389,7 +398,39 @@ public IEnumerable<TerminalTheme> GetPreInstalledThemes()
...
+                Author = "Canonical",
+                Name = "Ubuntu Bash",
+                PreInstalled = true,
+                Colors = new TerminalColors
+                {
+                    Foregroun = "#EEEEEE",
+                    Background = "#300A24",
+                    Cursor = "#CFF5DB",
+                    CursorAccent = "#000000",
+                    Selection = "rgba(207, 245, 219, 0.99609375)",
```
felixse: that's pretty much #f5f6f9. I think we ned to set alpha to something like 0.3-0.5 to make selected text still readable. This is a limitation of xterm.js, where we cannot set a selectedForeground color

felixse:
Thanks for your contribution 👍 

I think this will work fine, event if no wsl is installed it will only show a missing file error message if you try to launch it. Maybe we can add some special 'how to install wsl' message there in the future.

FluentTerminal.App/Services/Implementation/DefaultValueProvider.cs
```diff
@@ -389,7 +398,39 @@ public IEnumerable<TerminalTheme> GetPreInstalledThemes()
...
-            return new[] { defaultXterm, powerShell, homebrew, tomorrow };
+            var ubuntu = new TerminalTheme
+            {
+                Id = Guid.Parse("32519543-bf94-4db6-92a1-7bcec3966c82"),
+                Author = "Canonical",
+                Name = "Ubuntu Bash",
+                PreInstalled = true,
+                Colors = new TerminalColors
+                {
+                    Foregroun = "#EEEEEE",
```
notlmn: `Foreground`

Linard11:
@felixse would it be possible to check during installlation if the user actived the wsl windows component (check if `C:\windows\system32\wsl.exe` exists / `wsl `command is recognized) and add the configuration only then? Or maybe give the option to hide default profiles, so people who do not use is (or cmd/powershell for that matter) don't have to see it in the new tab dialogue

felixse:
I would check it on application startup, and then just hide it if wsl.exe is not found. But thats something we can do later on

FluentTerminal.App/Services/Implementation/DefaultValueProvider.cs
```diff
@@ -406,11 +406,11 @@ public IEnumerable<TerminalTheme> GetPreInstalledThemes()
                 PreInstalled = true,
                 Colors = new TerminalColors
                 {
-                    Foregroun = "#EEEEEE",
+                    Foreground = "#EEEEEE",
                     Background = "#300A24",
                     Cursor = "#CFF5DB",
                     CursorAccent = "#000000",
-                    Selection = "rgba(207, 245, 219, 0.99609375)",
+                    Selection = "4DABEDC1",
```
felixse: xterm.js does not support #aarrggbb, try rgba(207, 245, 219, 0.33) in this case

Linard11: Hmm, but just above, the homebrew theme uses it as well 🤔

felixse: I thought there was a problem, but maybe it was in a previous version of xterm.js, sorry😅

Linard11: you just merged while i pushed a commit 😓. Because I also noticed that i forgot the `#`

felixse: No worries, I will add it later today

Issue #37, opened 2018-06-10
Shortcuts to open a new tab with a given profile


Riebart:
The customized new-tab feature is great, but it would be even slicker to be able to shortcut that with a keyboard key combination to directly launch a new tab of the specific kind. Like ConEmu's tasks, being able to bind `Ctrl+Shift+P` to launch a new tab with my PowerShell profile would be perfect (with the default being my WSL, for example).

This feels like a nice simple one for me to get started on in a fork.

skipmeister123:
That does sound pretty easily to implement, but there might be conflicts
with custom profiles. For example, setting a shortcut to open a tab for a
custom profile, then removing said profile.

Hans Kokx

Riebart:
Fair point. Seems like a simple corner case to handle, but a good one to point out nonetheless.

skipmeister123:
Perhaps a good compromise would be a shortcut key specific to each profile?
When the profile is removed, the shortcut does nothing.

Hans Kokx

felixse:
Yes, I would also suggest to add the shortcut to the profile.

@Riebart  PRs are always welcome 😃

Riebart:
I'm working on this, getting close to having a PR ready.

felixse:
I did some bigger changes to architecture this weekend, let's hope this will merge well 💀 

Some comments on your code so far:
- I wouldn't reuse KeyBinding and Command here. Just create some similar model class that has an Guid for the shellProfile instead of the Command. 
- I think we can keep this a lot cleaner if we don't mix it up with the command keyBindings in general.
we could then just pass a collection of these ShellShortcutKeyBinding or whatever we name it to the view, and there just also loop through this collection and invoke something like openTabWithShellProfile(id) on the bridge.
- For the view I would just add it to the ShellProfiles page instead of the KeyBindings, I think this would be more logical, but correct me if I'm wrong here.
- When you pull the latest changes from master you might also see that I added the KeyBinding dialog to the IDialogService, so this should be easier to reuse now.

Oh and many thanks to your contributions to the documentation, I should have really elaborated this a bit more in detail 😄

Riebart:
There's some good discussion points there, thanks for the comments! I'll finish the last couple things, merge in master, fix that merge (which will inevitably break things), and then make a PR. We can discuss the architectural parts and implications on the overall design in the PR thread.

Riebart:
[...]

PR #45, opened 2018-06-18
Per shell keyboard shortcuts

Riebart:
There are several architectural discussions to have on this PR:

There are a couple breaking changes:

- This futzes with the GUIDs associated with the default shells
- Existing installs won't have the key bindings for the default shells, and I don't know how it'll respond to installs where there are custom shells without any key bindings.
- The command enum values have been explicitly set, so if they weren't set to their now explicit values, that'll cause issues in previous installs where it may just straight up crash on start.

Discussion points:

- Key bindings for shells are tied to the shell profile (for storage), and are saved with the shell profiles, not with the key bindings.
  - Key bindings for shells are still edited in the Key Bindings settings panel, and not with the shell profile itself. This was done mostly because I don't know XAML that well, and didn't want to futz around with that, but also because key bindings edit workflows don't follow the Edit/Save workflow of the shell profiles themselves (where they are read-only fields until you Edit the profile, etc...). This felt more natural from a UI/UX perspective, but did require some fudging to pass the ability to for the shell profile model to be able to trigger a key binding refresh on the key bindings panel when shells were added or deleted.
- This attempts to maintain the current flow of keyboard shortcut handling:
  - Register keyboard command handlers that call AddTerminal(), which had its signature changed to accept a shell profile to launch with. The registering of handlers that are available on startup is easy and handled in the MainViewModel initialization, and subsequent shell profiles are handled by having the MainViewModel add another KeyBindingChange event callback, which has access to the shells.
- Shells now have a KeyBinding property which is the list f bindings, as well as a KeyBindingCommand which is the Enum value of the command, basically taken as something in the 1024 and up range, or more specifically as something >= Command.ShellProfileShortcut.
  - There's an opportunity here to convert this KeyBindingCommand and the Id into some unified thing, but since the Command enum is used in a lot of places, conflating those two isn't something we can do without a major architectural decision, and as a PR contribution this didn't seem like the time for that.
- There's a conversation to be had about whether or not using the existing Command/KeyBinding/KeyboardCommandHandler paradigm is the right call for shell profiles, given that those shortcuts are always going to be caught by the TerminalView, and the action could just be taken there. That might simplify the code, but it would need some justification for why we're handling keyboard commands in a variety of different ways (and would make maintaining other keyboard shortcuts a little trickier, or at least more ambiguous).

Notes:

- There's a lot of linear searching through the shells list, which should be fixed by sorting out the discrepancy

### DIFF
--- a/FluentTerminal.App.Services/IKeyboardCommandService.cs
+++ b/FluentTerminal.App.Services/IKeyboardCommandService.cs
@@ -6,7 +6,7 @@ namespace FluentTerminal.App.Services
     public interface IKeyboardCommandService
     {
         void RegisterCommandHandler(Command command, Action handler);
-        void DeegisterCommandHandler(Command command);
+        void DeregisterCommandHandler(Command command);
         void SendCommand(Command command);
     }
 }
--- a/FluentTerminal.App.Services/IKeyboardCommandService.cs
+++ b/FluentTerminal.App.Services/IKeyboardCommandService.cs
@@ -6,6 +6,7 @@ namespace FluentTerminal.App.Services
     public interface IKeyboardCommandService
     {
         void RegisterCommandHandler(Command command, Action handler);
+        void DeegisterCommandHandler(Command command);
         void SendCommand(Command command);
     }
 }
--- a/FluentTerminal.App.Services/ISettingsService.cs
+++ b/FluentTerminal.App.Services/ISettingsService.cs
@@ -17,7 +17,7 @@ namespace FluentTerminal.App.Services
         void SaveDefaultShellProfileId(Guid id);
 
         IEnumerable<ShellProfile> GetShellProfiles();
-        void SaveShellProfile(ShellProfile shellProfile);
+        void SaveShellProfile(ShellProfile shellProfile, bool updateKeyBindings = false);
         void DeleteShellProfile(Guid id);
 
         TerminalOptions GetTerminalOptions();
--- a/FluentTerminal.App.Services/ISettingsService.cs
+++ b/FluentTerminal.App.Services/ISettingsService.cs
@@ -10,7 +10,7 @@ namespace FluentTerminal.App.Services
         event EventHandler<Guid> CurrentThemeChanged;
         event EventHandler<TerminalOptions> TerminalOptionsChanged;
         event EventHandler<ApplicationSettings> ApplicationSettingsChanged;
-        event EventHandler KeyBindingsChanged;
+        event EventHandler<Command?> KeyBindingsChanged;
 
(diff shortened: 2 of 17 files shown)


felixse:
I would really prefer going for a solution that does not rely on KeyBinding/Command. There is a lot of code added that just implicates that we are reusing something here that was not meant for this.
We could make KeyBinding a base class for both CommandKeyBinding (adds int Key) and ProfileKeyBinding (adds Guid ShellProfileId). Then we can just pass it into the webView and handle it pretty much like I described in #37 
This might add a second way of handling key shortcuts, but I think it is worth it from an architectural point of view. In the near future I would add some service to register keyBindings and taking care that no keybindings are used twice, this should deal with most of the issues we might have by having two systems in place.

If we go for ProfileKeyBinding we can give the model a an empty List<ProfileKeyBinding> as a default for both existing defaults and custom shells. That way we don't have to change any Ids of the default shells.

The XAML part is something I can take care of, should be no big deal. We can stick to the Edit/Save workflow by adding a readonly mode to the KeyBindingView. I can make a PR on your fork once we got there.

Riebart:
So a couple questions:

- Why do the shells have a GUID? Could an incrementing value be used, or is there a future plan here? I've always used UUIDs where I didn't have any synchronizing mechanism to guarantee uniqueness of IDs, but we have one of those here with the SettingsService.
- Which specific view do you see invoking the shell action? The Terminal View, or another one?

At the core of it, this is getting sticky I think because we're conflating a compile-time concept (a static list of actions we can invoke from the keyboard) with a runtime-concept (the list of shells we have available to invoke new tabs with). This implementation of shell shortcuts is definitely suboptimal, because I'm trying to use a system designed for the former for the latter, but I don't think it's the keybindings that need to change so much as the actions themselves. I may be wrong there.

felixse:
- I used GUIDs for the themes because they are stored in Roaming and might get merged if you use Fluent Terminal on multiple devices. When I got to implement ShellProfiles I just reused this to keep the implementation similar, but I agree int Ids would have worked as well.
- Basically the Javascript part which will then pass it down through the bridge to the TerminalView and the TerminalViewModel (pretty much like the existing keybindings, but in this case it should be enough to just pass the shell id)

Yes, ultimatively that's the problem. My point is that if we already store and handle them quite differently, we might also use some subtyping to better express their differentness. So yeah, the keybindings need to change, but we shouldn't change so much of the existing handling for the command keybindings.

Riebart:
This needs a major rewrite, given changes to the app since this initial implementation. I'm closing this, and will open a new one when I get to working on that feature.

Issue #42, opened 2018-06-16
Updater


skipmeister123:
There should be an updater mechanism, that doesn't require manual steps.  If getting into the Windows Store isn't possible, perhaps [WinSparkle](https://winsparkle.org/) is a solution?

felixse:
We would need an installer first if we want to use WinSparkle. I am still hoping that we can get in the Windows Store some time, so I would prefer not to invest much time into this.

How about checking for updates on startup and just showing a notification instead?

skipmeister123:
That'd be fine. Maybe a link to the releases page.

Hans Kokx

PR #46, opened 2018-06-20
Add notification for new version

M4xZ3r0:
Fix for Issue #42

### DIFF
--- a/FluentTerminal.SystemTray/Services/NotificationService.cs
+++ b/FluentTerminal.SystemTray/Services/NotificationService.cs
@@ -8,7 +8,7 @@ namespace FluentTerminal.SystemTray.Services
 {
 	public class NotificationService
 	{
-		public void ShowNotification(string title, string content)
+		public void ShowNotification(string title, string content, string url = null)
 		{
 			string xml = $@"<toast>
                             <visual>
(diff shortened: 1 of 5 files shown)


FluentTerminal.SystemTray/Services/NotificationService.cs
```diff
@@ -16,11 +19,30 @@ public void ShowNotification(string title, string content)
...
-            ToastNotificationManager.CreateToastNotifier().Show(toast);
-        }
-    }
+			var toast = new ToastNotification(doc);
+			ToastNotificationManager.CreateToastNotifier().Show(toast);
+		}
+
+		public void ShowNotification(string title, string content, string url)
+		{
+			string xml = $@"<toast>
```
felixse: May we can move this template to a string const instead of declaring it twice

FluentTerminal.SystemTray/Services/UpdateService.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using Windows.ApplicationModel;
+
+namespace FluentTerminal.SystemTray.Services
+{
+	public class UpdateService
+	{
+		private const string apiEndpoint = "https://api.github.com";
+
+		public bool GetLatestReleasedVersion()
```
felixse: From the name, I would expect this method to return some Version information object. Maybe rename to something like 'IsNewerVersionAvailable'

Riebart: Not sure if we should include it in this PR or not, but it would be nice to have the option to see if the newest version is a stable/beta/dev version, and optionally choose to update based on that. That might be a reason to return actual version strings or information.

M4xZ3r0: I actually wanted to use this api endpoint: https://developer.github.com/v3/repos/releases/#get-the-latest-release 
But there are only pre-releases currently, as such I don't get anything back from the endpoint.
Switching to this endpoint or including additional information in the message would not be a problem.

M4xZ3r0:
Fixed the comments from the review.
I would leave the check as it is as long as we have only pre-releases.

felixse:
👌

PR #49, opened 2018-06-21
Fix toggle window exception

cxfksword:
Current toggle window hotkey will throw an exception, fix it

### DIFF
--- a/FluentTerminal.SystemTray/Services/ToggleWindowService.cs
+++ b/FluentTerminal.SystemTray/Services/ToggleWindowService.cs
@@ -22,6 +22,8 @@ namespace FluentTerminal.SystemTray.Services
         private bool _disposedValue;
         private List<HotKey> _hotKeys;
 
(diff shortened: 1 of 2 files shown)


FluentTerminal.SystemTray/Services/ToggleWindowService.cs
```diff
@@ -145,8 +150,9 @@ private string GetActiveProcessFileName()
         private async void OnKeyPressed(object sender, KeyPressedEventArgs e)
         {
             if (GetActiveProcessFileName() == "Fluent Terminal")
-            {
-                SendKeys.Send("%{Tab}");
+            {
+                var hwnd = GetForegroundWindow();
+                ShowWindow(hwnd, SW_MINIMIZE);
```
felixse: That looks way better, thanks a lot 👍

felixse:
I will adapt the unit test on master so that we can get this merged

Issue #34, opened 2018-06-04
Any ETA when it will be available in Microsoft Store?


ghost:
P.S. Thanks a lot for the app! I have some ideas and I hope I'll contribute with pleasure when I get more free time 😉

felixse:
Current ETA is never 😄 

I would love to add this to the store, but there are 2 problems with getting this past the certification:

1. It is not compatible with Windows 10 S, since there is no cmd.exe/powershell.exe in this version of Windows, which would render the app completely useless.
2. I am bypassing a network isolation rule which blocks UWP apps from connecting to localhost. We need this for our websocket connection between the xterm.js in our WebView and our winpty backend.

We might be able to circumvent the second problem by switching to https://github.com/darrenstarr/VtNetCore.UWP once it is ready to use, but changes are high that the certification will still fail due to the Windows 10 S support

balcsida:
@felixse: Several, larger apps (ex Photoshop Express, Paint.NET) doesn't work on Windows 10 S either, I don't think this would be a problem.

ihodev:
> It is not compatible with Windows 10 S, since there is no cmd.exe/powershell.exe in this version of Windows, which would render the app completely useless.

As far as I can tell that's not a problem, because you can specify which Windows editions do you support by your app.

> I am bypassing a network isolation rule which blocks UWP apps from connecting to localhost. We need this for our websocket connection between the xterm.js in our WebView and our winpty backend.

Unfortunately I have nothing to say, but I hope that's can be implemented / fixed.

@felixse Thank you again, love your app 😉

balcsida:
> I am bypassing a network isolation rule which blocks UWP apps from connecting to localhost. We need this for our websocket connection between the xterm.js in our WebView and our winpty backend.

Edge extensions have the same limitations, which can be bypassed by the setting described on qzind/tray#6

There is a registry entry to that, so simply adding this reg file should resolve this:
```
Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Internet Settings\ZoneMap]
"IntranetName"=dword:00000000
```
See webextension-tools/webextension-toolbox#22 for more info (we use local WS connection to autoreload extension and tab on save)

felixse:
Hey,
we already have a workaround for this limitation (see: https://github.com/felixse/FluentTerminal/blob/master/FluentTerminal.SystemTray/LoopbackEnabler.cs)
[...]

PR #51, opened 2018-06-24
Update readme to address issues with loopback exemption

Riebart:
Updated the readme to provide instructions about enabling sideloading apps, and changing the wording around developer mode to be optional for _users_, at this point. Added the extra step to manually enable the loopback exemption if you've only got sideloading apps enabled.


felixse:
Thank you 👍 
We might as well write some installer script for this, but I would say we first see how #34 develops, maybe we can get this into the store soon enough

PR #68, opened 2018-08-18
Development environment documentation, and npm package bump

Riebart:
Broke the dev environment documentation out of the per-shell shortcuts (which need a serious rewrite after the large changes since then), and bumped npm package versions out of necessity (see 70e6dbf commit message).

Also adding a dummy development `Package.StoreAssociation.xml` that I copy-pasted and edited to build my test packages.


FluentTerminal.Client/package.json
```diff
@@ -10,9 +10,11 @@
   "author": "FS Apps",
   "license": "GPL-3.0",
   "devDependencies": {
-    "webpack": "^3.10.0"
+    "webpack": "^4.16.5",
+    "webpack-cli": "^3.1.0"
   },
   "dependencies": {
+    "add-dependency": "^1.1.0",
```
felixse: I'm not about this one. I don't think we need it, and if we need it I guess it should be in devDependencies

Riebart:
Yeah and I just noticed that while ncu reported version bumps it didn't actually commit then to the package files... Will fix

felixse:
Yes 👍

PR #70, opened 2018-08-18
Added support for Ctrl+<N> to switch to tab N.

Riebart:
Adding a handful of commands for switching to a specific tab number. This feature in Cmder was addicting, and I found myself constantly trying to use it in FT out of muscle memory.


felixse:
Looks good 👍

Issue #71, opened 2018-08-19
Looking for feedback: Tab-styling per theme


Riebart:
Been playing with what it will take to style tabs with the theme they are on (ties in with #28). Got a reasonable PoC done, but need a bit more work.

The way this works is that there will be a hanful of new values per theme, and the defaults (if unset) will be to use the current behaviour of following the Windows theme. The full list of colours that it could support is:

- Inactive tabs
  - Background
  - Foreground
  - Underline
  - Hover background
  - Hover foreground
  - Pressed background (foreground will match the hover foreground)
- Selected tab
  - Background
  - Foreground
  - Underline
  - Hover background (foreground be unchanged)
  - Pressed background (foreground be unchanged)

![image](https://user-images.githubusercontent.com/1278193/44304354-76f6ce80-a31f-11e8-9dd7-5b9310c2d1d8.png)

Thoughts? Suggestions? Things to do still:

- Convert the colour selection to match the existing grid style, since I was originally only planning to support a few colours, but with 11 it's time to move to just a box grid
- Need a good way of indicating in the UI via a ColourPicker that no colour has been chosen (the underlying value is `null`), so the system colours will be used.
  - This also needs a "clear colours" button for the tabs, to reset them back to the system defaults.

I'll update this issue when I push the code to my fork.

Riebart:
The idea is that you'd be able to leave them as windows default, or override. Leaving the values unset would let them follow the Windows default, and you'd be able to do that on a per-value basis.

The use case I have in mind is that I work on remote machines a lot, and this combined with the per-profile shortcuts that I need to rewrite would be excellent visual reminders of which tabs on which machines.

skipmeister123:
Ok, sure. I dig it.
--
Hans Kokx

Riebart:
Code pushed to https://github.com/Riebart/FluentTerminal/tree/tab_theme_colours. See the commits for specific TODO items remaining. Currently supports the default "leave it to Windows" behaviour, and overriding any parts separately of it with specific colours.

felixse:
Looks like a nice feature 😃 

Some comments:
- We could solve some problems by adding a Converter that returns the color if not null, or else looks up the desired default color (we can pass the resource key as ConverterParameter) from the resources. The problem I see right now is that we most likely break the default colors for light themes, and it would also keep or services and viewModels cleaner
- We should move 'underline selected tab' to the theme, just to keep it all in one place.
[...]

Issue #28, opened 2018-05-29
Default color scheme for profile


skipmeister123:
When creating a profile, I'd like to be able to choose a color theme for that profile.  For example, the Powershell theme should have the Powershell color scheme, while a custom profile may have a different one.

felixse:
I'm on this right now

PR #73, opened 2018-08-20
Tab theme colours

Riebart:
See #71 for details: This provides options to themes that allow you to set the colours of the tabs, including the background, foreground, underline, hovered-background, and other colours separately. Defaults, when unset, are to follow the system theme colours.

### Known issues
- Some colours don't work yet (likely attached to the wrong XAML UI element)
  - [ ] Inactive BG
  - [ ] Inactive click BG
  - [ ] Inactive hover FG
  - [ ] Active click BG
- [x] "Colour" is used in some locations, so a global search-replace for "Color" is necessary
- Naming of a couple items:
  - [x] `TabHoverBackground` should be `TabHoverBackgroundColor`
  - [ ] `ColorTupleToBrushConverter` doesn't actually take in a Tuple
- [x] `ColorTupleToBrushConverter` takes in a weird string passed from the output of properties in the TerminalViewModel intended to convey context and intention to the convreter. There's probably a cleaner way of doing this.
- [ ] The inactive tab system hover colour adjustment doesn't work (the tab background should light up/darken a bit when hovered, but this no longer happens even with default themes with no tab colour overrides)
- [x] A sister option to "underline selected tab" being "underline inactive tabs too". This factors into the `IsUnderlined` logic, as currently "Underline selected tab" is being abused, and causes all tabs to be underlined.
  - The solution I chose is simpler than another option: inactive tabs are only underlined if they specify an inactive tab underline colour.
- [ ] There is no way to clear a picked colour back to `null`, and thus return the behaviour back to system default.
- [ ] I've made no efforts yet to check/implement import/export of these new properties.


Riebart:
Some screenshots of things in action:

The new theme colour-selection UI, with none of the colours chosen.
image

The default xterm.js theme, obeying system default behaviour for tab colours. Note the issue with underlining inactive tabs due to a current mistreatment of the underlining config option.
image

An active and an inactive tab, note that some of the colours don't work, as noted in the PR ToDo list.
image

Just like the default behaviour, the inactive tab background is transparent.
image

Tab foregrounds, if not specified, use the same dynamic light/dark theming for selecting the foreground text based on the effective tab background colour. If transparent, this is based on the theme background, otherwise it is based on any set tab background colour.
image

felixse:
Hey. I'm somewhat busy with other stuff right now, but I had time to checkout your changes yesterday and play around with it. One problem I see right now is that the new template is pretty lacking in terms of nice hover/reveal effects etc. 😅
[This was a pretty interesting read.](http://loekvandenouweland.com/content/UWP-lightweight-listview-styling.html) It would be cool if we could set the resources of a ListViewItem programmatically. Or otherwise copy it and make sure to only change the parts where these reource keys were used. I hope a find some time this weekend to investigate this.

In regards of where to save this I'm still not convinced that themes is the right place. It would be a pretty patchy way to use it this way: Prepare themes with according tab colors, then set these theme for each profile. I guess some predefined tab colors to select for your profiles would be easier. That way we can also over some temporary coloring in the context menu of a tab

Riebart:
Yeah, I suspect that many of the brushes are actually the schmancy new Acrylic brushes (which we'll be seeing more of in 1809 in the standard Windows UI too), and the `SolidColorBrush` that I'm rendering them as. The simple answer would be to, instead of just fetching the _Color_ attribute of the brush, create a new brush of the same type (or just use the brush directly from the Resource dict for system colours), and overwrite only the Color (and optionally the opacity) attributes. This could pretty simply be handled in the Converter that generates the brushes, once I figure out what the structure of the default brushes is.(the `AcrylicBrush` class looks like it can have all sorts of composition brushes, so that could get entertaining).

As for where to store them, that's six of one and half dozen of the other. Depends on which workflow we want to optimize for:

- If people just want to customize their themes to also include tab colour schemes, then having them in the theme makes the most sense, plus shares the semantics of setting all of the colours in one place.
- If people want to have tab colours per profile, and we don't want to add a Clone action (we can do it now with export/import, but that's clunky) to themes (since "Create new" always seems to create one based on xterm.js), then adding it to the profile makes the most sense.
  - Supporting #28 which would asks for a per-profile theme anyway means that adding the Clone action to a theme, and keeping the tab colours in the theme.

It doesn't really matter, since moving the UI elements to the shell and changing the `GetCurrentTheme()` calls to the shell equivalent is easy.

felixse:
Maybe the style you copied was from an older sdk version? When I played around with the resource keys I also used solid colors, but effect were still present.

I think going for the color tagging workflow helps more users in the long run, and it would also fit your needs, right?

Riebart:
Quite possible that I'm just using the wrong style description. I'm out of town for a bit, so I won't be able to dig into this, but the reference I used is noted in a comment in `TabBar.xaml` ([Ref](https://msdn.microsoft.com/en-us/library/windows/apps/mt299136.aspx)). I used the simple one (i.e. that used when the item is an `ItemsStackPanel`), not the complicated one, so that's likely the source of the issue. It should be a simple test to copy-paste in the different defaults, comment out my bindings, and see which produces the correct default behaviour.

I mean, really, we can have these in _both_ places, if we want. The UI config elements are simple, and it would be a natural to have them in the theme, but if specified in the profile, then the profile takes over. In any event I am generally indifferent to specific location, and I'll defer to your call on this. 😸

felixse:
Profile edit is kind of broken right now (quite sure it was already broken before we started this), but apart from that it's mostly fine now. Please have a look and tell me what you think 😄

Riebart:
Looks great! I took a crack at adding a global underline-instead-of-background-colour option for a slightly more subtle visual hint. It's a general setting, since I like how slick your tab theme colour selection is, and choosing underlining behaviour and theme colour would have messed that up.

felixse:
That's a good idea, having every tab with a background color can get quite noisy. I would say once we the broken profile editing this is good to merge

Riebart:
By broken profile editing, do you mean persistent changes to the default profiles, or something else? I'd be happy to take a look at that this weekend, I'll just need to know what to fix. :)

felixse:
Cancel is broken somehow, except for just created profiles, also some crashes when saving because the color profile was not set correctly. I guess this is some compiled binding weirdness, maybe switching to Binding helps.

For the new 'underline instead of background color' option I would refactor it to a radio button group, this will make it more expressive I guess (Like Inactive colored tab style: Underline/Colored)

felixse:
Looks good to me now, if you haven't any further points I would merge this

Riebart:
Looks great! 😸

Issue #82, opened 2018-09-21
This is pretty much a nightmare for non-Windows developers!


RichardBronosky:
I don't think I'm alone in this because of WSL. There are going to be a lot of people thinking maybe they can try Windows, but there is no decent terminal. Seriously, none. But here is what a Linux or Mac refuge is going to experience:

First:
`git clone https://github.com/felixse/FluentTerminal.git`

Then follow...
https://github.com/felixse/FluentTerminal#how-to-install
Hmmm, `find . -name '*cer'` and `find . -name '*appx*'` reveal nothing.

Okay, follow...
https://github.com/felixse/FluentTerminal#how-to-build
* `choco install nodejs` works
* `choco install VisualStudio2017Community` works
* Developer Mode enabled

On to...
https://github.com/felixse/FluentTerminal#visual-studio-2017-configuration
* Uh, "Under Workloads"? Wat?
* "Under individual components"? Wat?
* Maybe I'll just try to open the FluentTerminal.sln file...
![screenshot_15](https://user-images.githubusercontent.com/13620/45908087-b86c1500-bdc8-11e8-8b90-2dcc5f263669.png)
* Ok, I guess...
![screenshot_16](https://user-images.githubusercontent.com/13620/45908118-d2a5f300-bdc8-11e8-9d8d-0b371782a4ff.png)
* Oh, good. It's going to install them for me...
![screenshot_17](https://user-images.githubusercontent.com/13620/45908144-03862800-bdc9-11e8-9551-3c1ccebcb40a.png)
* Wat?

Hmmm... Maybe the new MacBook Pro isn't as terrible as I thought.

Hey, it's your project and you are free to do it however you want. I just know that as the maintainer of many projects myself, I always wonder what the experience is like for people not on the inside. I'd love feedback like this on https://github.com/RichardBronosky/pi-init2 and I go and give presentations and hands-on workshops at [Meetups](https://www.meetup.com/Decatur-Makers/events/llsgmnyxkbmb/) to get it.

skipmeister123:
Why not just download the latest release off the releases page? 🤔
--
Hans Kokx

RichardBronosky:
Also, this is what I get when I open the `.cer` and allow it to "Automatically select the certificate store...". I would suggest some additional instruction here too.
![screenshot_18](https://user-images.githubusercontent.com/13620/45908838-970d2800-bdcc-11e8-806d-220d48b720c4.png)

benplm:
[...]

PR #83, opened 2018-09-28
Clarifying installing and dev environment setup instructions.

Riebart:
Working to address some of the comments in #82.

Note that sideloading apps on Windows, like Linux, is the nonstandard way to get stuff. The goal is to get this into the Microsoft Store, at which point the install steps, in their entirety, are:

> Install from the Microsoft Store.

What you're doing is installing a development version/bleeding edge, that has yet to make it into any repositories. Compare this with installing a Python package with C dependencies that require compilation from a git repo's `master`.


README.md
```diff
@@ -31,21 +31,27 @@ A Terminal Emulator based on UWP and web technologies.
...
 - Grab the [latest NodeJS 8.x installer](https://nodejs.org/en/download/), and install ensuring that both NPM and "install into system PATH" are checked during the install process.
 - Grab the [latest Visual Studio 2017 Community installer](https://www.visualstudio.com/downloads/) and install (See next section)
 - [Enable developer mode in Windows 10](https://docs.microsoft.com/en-US/windows/uwp/get-started/enable-your-device-for-development)
 
-### Visual Studio 2017 Configuration
+### Installing Visual Studio 2017
+
+There's some specific parts of VS2017 that we need to make sure are available when you go through the install process.
 
 - Under "Workloads" select:
```
felixse: We should add something link 'in Visual Studio Installer', since you might also install it using chocolatey without having the Installer open

README.md
```diff
@@ -31,21 +31,27 @@ A Terminal Emulator based on UWP and web technologies.
...
 - Split screen support
 
-## How to install
+## How to install (as an end-user)
 
+- Download the latest [release](https://github.com/felixse/FluentTerminal/releases)
 - [Enable sideloading apps](https://www.windowscentral.com/how-enable-windows-10-sideload-apps-outside-store)
   - Alternatively, [enable developer mode](https://docs.microsoft.com/en-US/windows/uwp/get-started/enable-your-device-for-development) if you plan to do UWP app development. **For most users that do not intend to do UWP app development, developer mode will not be necessary.**
 - Install the *.cer file into `Local Machine` -> `Trusted Root Certification Authorities`
+  - This will require administrator elevation. If you installed the certificate and did not have to elevate, you likely installed it into an incorrect certificate store.
```
felixse: maybe add something like 'do not trust the doubleclick'? 😄

Riebart: Oh, yeah, good one. I'll add something specific. Think we need screenshots?

README.md
Riebart: Ah, right. I've personally never used Chocolatey, so I'll add a specific note along the lines of "Don't install with chocolatey unless you're sure you can do this, use the installer". Would screenshots help?

Riebart: I added some screenshots to make it pretty clear. I think that's a bit more concise than trying to explain it.

README.md
Riebart: I added a blurb that automated installers should be avoided to ensure they get a chance to choose all of the correct installation options.

PR #85, opened 2018-09-30
Per-shell-profile keyboard shortcuts

Riebart:
Architected better this time.

image

image

- The `Command` enum is left alone, but there is now a clear distinction between a command and a key binding. Specifically, a new class, `AbstractCommand`, is used everywhere, which supports `EnumCommand` as well as others. Specifically this implements the necessary parts to be hashable and usable as Dictionary keys.
  - This manifests in the KeyBinding type no longer having a Command property.
- Whenever xterm.js sends back a command, the string is passed through the handlers until it reaches a point where it can be resolved into a proper `AbstractCommand` by the `SettingsService` instance. From then on the AbstractCommand is used.
- When shell profiles are created or deleted, there is an event handler that alerts all of the MainViewModel instances, which handle updating their keyboard command handlers.
- For the UI elements, I modified the KeyBindingView and associated model to support an `Editable` property that can be set at creation time. This has the effect of hiding (when false) the Edit and Delete actions.
- The ShellProfile Settings/ViewModel use the KeyBindingsView stuff, and another button in the editable case for adding additional keyboard shortcuts to profiles. These are committed and the `KeyBindingsChanged` event is raised whenever a shell profile is saved.
- Since `KeyBinding`s no longer have a Command property, `FlattenKeyBindings()` now returns a `Dictionary<string, object>` that includes all of the same keys and values as before, such that it JSON encodes the same and quacks the same as the old KeyBinding type.
- VS wanted to correct a few files with inconsistent line endings to CRLF, so that happened.

With this architecture, adding additional commands, not sure what they'd be, is a much lighter lifting operation:

- Create a new subclass of `AbstractCommand`
- Add parsing capability to `SettingsService.ParseCommandString` to convert the `ToString()` output of your new command class back into a command class.
- Make sure the MainViewModel instances can add the new commands to their handlers on create, and on events where the commands may change.

### ToDo

- The EnumCommand needs to use `EnumHelper.GetEnumDescription()` for the `Description` property, but because of where that class sits it causes issues.
  - Proposed solution: Refactor that into a Utilities project, and depend on that, which will simplify dependencies. The hacky solution involves copy paste.
- When editing the shell profile, the keyboard shortcut "name" text doesn't update as the shell profile name is updated, and in fact won't update until the Settings window is closed and re-opened.
  - Proposed solution: This could be alleviated by making the `CommandName` property of `KeyBindingsViewModel` not readonly for public setting, and or providing an optional override in the constructor so we can use a generic name in that specific case.


felixse:
This sounds awesome 😃 
I will check this out in detail tonight (CEST)

My thoughts on the todos:

- We could use a converter, then we only need the helper in the viewModels (which sounds good, because the model shouldn't care about its readability).
- Mutable CommandName is no problem

Riebart:
All tests are passing, the second ToDo was obviated by just not putting the shell profile name in that text box (the text was getting too long for that part of the UI anyway), and the first one isn't that bad, and we only need a couple lines of borrowed code.

I've been using this as my daily driver for a few weeks now, and it's stable and works as expected. :)

felixse:
I get a crash when I try to open settings.
Line 55 in ShellProfileViewModel is passing shellProfile.KeyBinding, which is null in my case

Riebart:
Pushed a change, but I can't seem to test it, since "Uninstall" doesn't actually clear all of the data for the app, so I'm having trouble testing this migration path.

FluentTerminal.Models/KeyBinding.cs
```diff
@@ -1,14 +1,28 @@
...
     public class KeyBinding
     {
-        public Command Command { get; set; }
         public int Key { get; set; }
         public bool Ctrl { get; set; }
         public bool Alt { get; set; }
         public bool Shift { get; set; }
         public bool Meta { get; set; }
+
+        public Dictionary<string, object> ToDict(AbstractCommand command)
```
felixse: this seems odd, why does the keybinding know about commands? I think it would be easier to just change the Command property to string, then we can check if this string is either a guid of a profile or a string presentation of one of the command enums values.

FluentTerminal.Models/AbstractCommand.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        public static bool operator !=(AbstractCommand A, AbstractCommand B)
+        {
+            return !A.Equals(B);
+        }
+
+        public static implicit operator AbstractCommand(Command input)
+        {
+            return new EnumCommand<Command>(input);
```
felixse: It looks really interesting, but I guess its breaking the abstracting if the abstract class nows about its derived classes

FluentTerminal.Models/ShellProfile.cs
```diff
@@ -11,5 +12,6 @@ public class ShellProfile
         public string Location { get; set; }
         public string WorkingDirectory { get; set; }
         public int TabThemeId { get; set; }
+        public ICollection<KeyBinding> KeyBinding { get; set; }
```
felixse: Should be called 'KeyBindings' to emphasize that it is a collection

FluentTerminal.App.Services/Implementation/SettingsService.cs
```diff
@@ -52,10 +52,16 @@ public SettingsService(IDefaultValueProvider defaultValueProvider, ApplicationDa
 
         public event EventHandler KeyBindingsChanged;
 
+        /// <summary>
+        /// Indicate that a change has occurred to the shell profile collection. If a profile was added, the furst tuple value will be true, and if a profile was removed it will be false.
+        /// </summary>
+        public event EventHandler<Tuple<bool, ShellProfile>> ShellProfileCollectionChanged;
```
felixse: Why don't we split it into ShellProfileAdded and ShellProfileDeleted? Would be more expressive

FluentTerminal.App/Views/KeyBindingsView.xaml.cs
```diff
@@ -9,15 +9,23 @@ public sealed partial class KeyBindingsView : UserControl
         public static readonly DependencyProperty ViewModelProperty =
                     DependencyProperty.Register(nameof(ViewModel), typeof(KeyBindingsViewModel), typeof(KeyBindingsView), new PropertyMetadata(null));
 
+        private bool _editable = true;
```
felixse: under what circumstances would this be used instead of the viewModel property?

FluentTerminal.Models/AbstractCommand.cs
felixse: I can have a look at it this weekend

felixse:
Sorry for being so picky, I think we are on a good direction here 😄

FluentTerminal.Models/KeyBinding.cs
Riebart: Problem is that the command was previously embedded in the KeyBinding object (https://github.com/felixse/FluentTerminal/blob/master/FluentTerminal.Models/KeyBinding.cs), but isn't anymore. It used to just serialize the keybindings list, which included the commands. Since the keybindings don't include commands anymore, we need something that adds it back in.

felixse: we could add `public string Command { get; set; }` which is either Command.ToString() or shellProfile.Id.ToString()

PR #86, opened 2018-10-01
Return an empty string when attempting to paste non-text clipboard buffer

Riebart:
Returning null Task reference was causing a crash when attempting to paste a non-text buffer.

### DIFF
--- a/FluentTerminal.App/Services/ClipboardService.cs
+++ b/FluentTerminal.App/Services/ClipboardService.cs
@@ -13,7 +13,8 @@ namespace FluentTerminal.App.Services
             {
                 return content.GetTextAsync().AsTask();
             }
-            return null;
+            // Otherwise return a new task that just sends an empty string.
+            return new Task<string>(new Func<string>(() => ""));
         }
 
         public void SetText(string text)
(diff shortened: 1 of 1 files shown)


FluentTerminal.App/Services/ClipboardService.cs
```diff
@@ -13,7 +13,8 @@ public Task<string> GetText()
             {
                 return content.GetTextAsync().AsTask();
             }
-            return null;
+            // Otherwise return a new task that just sends an empty string.
+            return new Task<string>(new Func<string>(() => ""));
```
felixse: [Returning Task.FromResult(string.Empty) might be easier](https://msdn.microsoft.com/en-us/library/windows/apps/hh194922(v=vs.105).aspx)

felixse:
Yeah, I had this happening once and forget to track it, thank you 👍

FluentTerminal.App/Services/ClipboardService.cs
Riebart: Quite possibly. This was yoinked from StackOverflow because modern .NET isn't my forte. 😸 

I'll change it.

PR #92, opened 2018-10-08
Add automatic install script

justingallagher:
Add a script based off of Visual Studio's `Add-AppDevPackage.ps1`, which walks the user through enabling sideloading, installing the certificate and package, and activating the Explorer integration. This should help alleviate concerns from users that the install process is too complex (#82).

I added the instructions to the README. The manual install instructions are still present, in case anyone has issues with the script.

These are the major changes to the original script:

* A windows developer license is no longer required, only enabling sideloading. See `CheckIfNeedToEnableSideloading`.
* At the end of the install process, the script asks the user if they want to install the context menu integration (defaulting to No). See `ConfirmContextMenuInstall` and `InstallContextMenuIntegration`.
* UI Strings are stored in a local class (`UIStrings`), rather than using localized resources. Since I had to modify strings for our purposes, we would need new translations of many of them.


felixse:
This is awesome, thank you very much 😃
I will add this to the release package starting with 0.2.1.0

PR #115, opened 2018-11-06
Fix display issue when settings window is too narrow

ericcornelissen:
Disable compact mode for the settings window (which removes the sidebar) entirely. Instead now the side bar is just collapsed at `560` (I guess pixels?).

### Preview

settings-window-fix-preview


felixse:
Thanks 👍

PR #124, opened 2018-11-12
added missing test for KeyboardCommandService

Meir017:
added missing test for KeyboardCommandService for Command.ToggleWindow


felixse:
Thank you 👍

Issue #153, opened 2018-12-17
Configurable padding


fauzie811:
I don't like padding on terminal windows. It would be nice to have an option to configure the padding.

Thank you.

PR #155, opened 2018-12-17
Add Padding setting

fauzie811:
#153

### DIFF
--- a/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
+++ b/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
@@ -305,6 +305,7 @@ namespace FluentTerminal.App.Services.Implementation
                 FontFamily = "Consolas",
                 FontSize = 13,
                 BackgroundOpacity = 0.8,
+                Padding: 12,
                 ScrollBackLimit = 1000
             };
         }
--- a/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
+++ b/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
@@ -305,7 +305,7 @@ namespace FluentTerminal.App.Services.Implementation
                 FontFamily = "Consolas",
                 FontSize = 13,
                 BackgroundOpacity = 0.8,
-                Padding: 12,
+                Padding = 12,
                 ScrollBackLimit = 1000
             };
         }
--- a/FluentTerminal.App/Views/SettingsPages/TerminalSettings.xaml
+++ b/FluentTerminal.App/Views/SettingsPages/TerminalSettings.xaml
@@ -57,6 +57,15 @@
                     Minimum="0"
                     StepFrequency="0.01"
                     Value="{x:Bind ViewModel.BackgroundOpacity, Mode=TwoWay}" />
+                <Slider
+                    Width="200"
+                    Margin="0,0,0,24"
+                    HorizontalAlignment="Left"
+                    Header="Padding"
+                    Maximum="100"
+                    Minimum="0"
+                    StepFrequency="1"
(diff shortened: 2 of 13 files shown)


FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
```diff
@@ -305,6 +305,7 @@ public TerminalOptions GetDefaultTerminalOptions()
                 FontFamily = "Consolas",
                 FontSize = 13,
                 BackgroundOpacity = 0.8,
+                Padding: 12,
```
ericcornelissen: Not sure if and why it worked for you, but I think this should be an `=` instead of a `:`

```suggestion
                Padding = 12,
```

FluentTerminal.App/Views/SettingsPages/TerminalSettings.xaml
```diff
@@ -57,6 +57,15 @@
                     Minimum="0"
                     StepFrequency="0.01"
                     Value="{x:Bind ViewModel.BackgroundOpacity, Mode=TwoWay}" />
+                <Slider
+                    Width="200"
+                    Margin="0,0,0,24"
+                    HorizontalAlignment="Left"
+                    Header="Padding"
+                    Maximum="100"
```
ericcornelissen: Although it is very subjective, I don't think anyone would want a padding of `100px`. Personally I would set this to something in the range of `30` (perhaps somewhat low) to `50` (already very high).

fauzie811:
Sorry, my bad. I forgot to push the latest commit.

FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
fauzie811: Sorry, my bad. I forgot to push the latest commit.

fauzie811:
To be honest. I haven't compiled the code myself, because I couldn't. I haven't finished downloading the Windows 10 SDK. I'm so sorry, I just couldn't wait.

In theory, it should work now.

FluentTerminal.App/Views/SettingsPages/TerminalSettings.xaml
fauzie811: I agree. I think 30 is enough.

ericcornelissen:
Haha okay, no problem (perhaps say that in the PR description next time 😉). I also figured out why it wasn't working for me, simply because I didn't rebuild the script 😅 

The problem I'm having right now is that when the padding is set to a value higher then `12px` it pushes the `.terminal` outside the limits of the container. I imagine this can be fixed by using [box-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing) but it didn't work upon my first and only try 😞

fauzie811:
This time I'm 100% sure it'll work. 🤞


Maybe. 🤣

FluentTerminal.Client/src/index.js
```diff
@@ -56,8 +56,10 @@ function createTerminal(options, theme, keyBindings) {
 
   term.open(terminalContainer);
   term.winptyCompatInit();
-  term.fit();
   term.focus();
+  
+  setPadding(options.padding);
+  term.fit();
```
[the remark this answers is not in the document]
fauzie811: Huh, how come I didn't notice that? 🤔

ericcornelissen:
I will just leave this preview I made here for anyone interested. Note that the visual bugs regarding the text is caused by xterm's `term.fit()`.

padding_preview

felixse:
Looks great @fauzie811 👍 
And also thanks for reviewing this @ericcornelissen 

> After that its up to @felixse to decide on the padding value range and merge the PR 😃
Lets make it 64px 🤔

PR #171, opened 2019-01-10
fix #114 scrolling issue on presicion touchpads

### LINKED ISSUE #114: Touchpad scrolling not working
Scrolling with mouse works but scrolling with touchpad doesn't. The entire content container gets pulled, not the content inside it.

MikeKovarik:
Hello,
first of all - amazing app, what I've been longing to find or wanting to make myself for a long time.

Since Surface Pro is my mainly driver, the precision touhpad scrolling issue #114 has been a dealbreaker for a while now so I decided to fix it. I've been making some EdgeHTML & HTML UWP apps for a while so I already came across this is pretty basic bug (*unless MS decided to market it a feature and call it a day :D*) plaguing all my projects.

**What's been happening?**
Scroll events are getting captured by `<body>` or `<html>` and since it has no scrollbar and no content larger than elements visible portion, EdgeHTML does the overflow bounce effect.

There's also adjacent issue to the scrolling which only surfaces on precision touchpads - pinch zooming. Also side scrolling.

**What's I've done**
* Applied overflow hidden on body and html making it not scrollable because there's no content to scroll
* Made the xterm's container scrollable and threw in pointer events for a good measure
* prevented zooming, both with touchpad and through touch.
* bonus: made the scrollbar autohide because it's super cool ;)

**Observations**
Unfortunately the scrolling isn't smooth as one would expect from UWP app because the scrolling of the terminal is handled by xterm, making it jump an entire line every time you scroll.

Touch scrolling is does not work, instead touch creates text selection. I did not investigate further. I suppose it's also on the xterm's side of thing. And maybe it's a desired behavior. If not, we could create separate issue and fix it later.

This may not be the complete solution, though it's a step in the right direction. I've been testing it for a while now and it works great :)

### DIFF
--- a/FluentTerminal.Client/src/style.css
+++ b/FluentTerminal.Client/src/style.css
@@ -1,15 +1,24 @@
 * {
     margin: 0;
+    -ms-overflow-style: -ms-autohiding-scrollbar;
 }
 
 html, body {
     background: transparent;
     height: 100%;
-  }
+    width: 100%;
+    overflow: hidden;
+    -ms-content-zooming: none;
+}
 
 #terminal-container {
     width: 100%;
     height: 100%;
+    /* reenable scrolling only in the exterm */
+    overflow-y: auto;
+    overflow-x: hidden;
+    touch-action: pan-y;
+    -ms-touch-action: pan-y;
 }
 
 .xterm .xterm-viewport {
--- a/FluentTerminal.Client/src/style.css
+++ b/FluentTerminal.Client/src/style.css
@@ -1,6 +1,5 @@
 * {
     margin: 0;
-    -ms-overflow-style: -ms-autohiding-scrollbar;
 }
 
 html, body {
--- a/FluentTerminal.Client/src/style.css
+++ b/FluentTerminal.Client/src/style.css
@@ -14,7 +14,7 @@ html, body {
 #terminal-container {
     width: 100%;
     height: 100%;
-    /* reenable scrolling only in the exterm */
+    /* re-enable scrolling only in the xterm container */
     overflow-y: auto;
     overflow-x: hidden;
     touch-action: pan-y;
--- a/FluentTerminal.Client/src/index.html
+++ b/FluentTerminal.Client/src/index.html
@@ -5,6 +5,7 @@
     <meta charset="UTF-8">
     <link rel="stylesheet" href="xterm.css" />
     <link rel="stylesheet" href="style.css" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no">
 </head>
 
 <body>
--- a/FluentTerminal.Client/src/index.js
+++ b/FluentTerminal.Client/src/index.js
@@ -145,14 +145,11 @@ function changeOptions(options) {
 }
 
 function setScrollBarStyle(scrollBarStyle) {
-  if (scrollBarStyle == 'hidden') {
-    document.getElementById('terminal-container').style['-ms-overflow-style'] = 'none';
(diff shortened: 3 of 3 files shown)


skipmeister123:
😮 thank you!
--
Hans Kokx

ericcornelissen:
Works like a charm 👍 

> Unfortunately the scrolling isn't smooth as one would expect from UWP app because the scrolling of the terminal is handled by xterm, making it jump an entire line every time you scroll.

I don't notice any difference in the smoothness of the scrolling between `felixse:master` and `MikeKovarik:master`.

MikeKovarik:
@ericcornelissen I rephrased the comment.

What i meant with the comment on scrolling was not that my PR would cause regression. Instead I sighed that the scrolling was never smooth.

To explain further. UWP apps and Edge has the amazing buttery smooth scrolling that moves the content pixel by pixel and also has the inertia when you flick the fingers on touchpad. As compared to for example chrome on windows (up until like a year ago) which always jumped like 40+ pixels at a time instead of animating to the position. Not only is it nice and fancy but it gives visual queue to the user "oh it's going from here to there", just like any good transition in mobile apps, as opposed the the typical "oh, the screen flashed and now my content is gone, where is it? up, or down? how do I go back?."

The moment I first opened Fluent Terminal i hoped it would have a smooth scrolling since it's powered by EdgeHTML. But instead scrolling is (~I believe, did not look into it further~) handled by xterm so that it only always scrolls exactly one line at a time. e.g. jumps to the next line by the 18+- pixels at a time.

note: found it, it's being discussed here https://github.com/xtermjs/xterm.js/issues/1140

FluentTerminal.Client/src/style.css
```diff
@@ -1,15 +1,24 @@
 * {
     margin: 0;
+    -ms-overflow-style: -ms-autohiding-scrollbar;
```
felixse: Can you try to make this configurable again? https://github.com/felixse/FluentTerminal/blob/8131140cb0f358826e46ebcc69cb4cec2c971598/FluentTerminal.Client/src/index.js#L147

felixse:
Looks great, just one minor issue before we can merge this 👍

MikeKovarik:
alright. i didn't notice this was an option. Removed that and simplified related js code

ericcornelissen:
Considering the JS code simplifications (but unrelated to the PR at large) I would suggest, from a software engineering point of view, to 1) add a default value and 2) convert the string `'-ms-overflow-style'` into a variable or even move `terminalContainer.style['-ms-overflow-style'] = 'none'` inside a function/outside the switch-statement. For example (in the spirit of @MikeKovarik's changes):

```javascript
function setScrollBarStyle(scrollBarStyle) {
	var DEFAULT_OVERFLOW_STYLE = 'scrollbar'; // I'm not sure what the actual default value is
	var setOverflowStyle = (value) => terminalContainer.style['-ms-overflow-style'] = value; // This prevents one from having to update 3 (or 4) lines when e.g. the string '-ms-overflow-style' needs to be changed

	switch (scrollBarStyle) {
	  	case 'hidden':     return setOverflowStyle('none');
		case 'autoHiding': return setOverflowStyle('-ms-autohiding-scrollbar');
		case 'visible':    return setOverflowStyle('scrollbar');
		default:           return setOverflowStyle(DEFAULT_OVERFLOW_STYLE); 
	}
}
```

felixse:
@ericcornelissen You're right, but I'm quite optimistic that `-ms-overflow-style` won't change anytime soon 😃

Issue #110, opened 2018-10-28
Line breaks


sinnlosername:
I got some trouble with line breaks. Bash on Windows (WSL) works with LF as it is a linux system, while the windows cmd works (afaik)´with CRLF.

If I copy a text from notepad with CRLF and paste it into fluentterm=>WSL it will do twice as much new lines. So for example

```
test1
test2
test3
```

will be pasted as 

```
test1

test2

test3

```

It would be nice to specify a linebreak per profile and convert pasted text to it. So if I specify "LF" for WSL and paste a text with "CRLF" it will replace all CRLF with LF.

skipmeister123:
I see this when pasting into vim, too.
--
Hans Kokx

Riebart:
I have this built in a branch, I could create a PR, and divorce it from my tab shortcut code.

felixse:
@Riebart thanks for taking care of this 😄  I don't mind if it is in the same PR as tab shortcuts. How is your progress over there?

Riebart:
Been crazy, haven't had a free evening in three weeks. But a bunch of stuff finished up this weekend, I should be able to polish up that pr this week,  thanks for you patience!

felixse:
No worries, take your time 😃

Riebart:
Ok I've got the shell shortcuts PR updated and passing tests. I'll put in the line breaks PR once that's accepted, since it builds on some of the changes to the profile settings page in the shell shortcuts PR.

My line break editing PR also comes with a new keyboard command, "Paste without newlines" which I found really helpful in Cmder, since the copy-paste wasn't always aware of when what I was copying was soft wrapping at the terminal edge, and would occasionally insert spurious line breaks. Also useful when copying from Powershell, as those are regularly hard-wrapped in the copied text.

skipmeister123:
As long as you're running a relatively recent build, you'll get an update notification.  That's something that was implemented a couple builds ago.

--
Hans Kokx
On 11/7/2018 12:24:00 PM, Florian <notifications@github.com> wrote:
By the way, how do I know when a new update is available? Is there an update notification or do I need to check them manually?

sinnlosername:
any updates

sinnlosername:
Any news?

Riebart:
Taking me a bit longer to rebase the changes, some of the code blocks have changed. Should be in this week. Sorry for the delay. 😸

Riebart:
[...]

Issue #169, opened 2019-01-08
Copy multiline text may fail for some lines


thochra:
Hello.

Context: Wsl / ssh to remote linux and tailed some logs

When the displayed line in the terminal ends with a conventional character the copy works well.
But, if the line ends with a space character then the log line is splitted.
Check the red square on the next picture and compare with the result pasted in vscode.

![image](https://user-images.githubusercontent.com/11785076/50863063-c5276f80-139d-11e9-957d-96ee0a65c08a.png)

As we can see with the previous log line, if the displayed line starts with a space, the complete log line is kept.

I noticed they handled it in vscode's terminal which is also using xterm.

Could it be possible the simple use of their version of xterm would solve the issue?
https://github.com/Microsoft/vscode/blob/1.30.2/package.json
https://www.npmjs.com/package/vscode-xterm

In the meantime I wonder if vscode-xterm is not too far from xterm, because I did cross read this post https://codeburst.io/source-reading-how-is-new-terminal-in-vs-code-so-fast-10a40f7f8792.

Aside that, nice work! Looks promising 👍

thochra:
Hello, I tested again with version 0.4.1.0.
The line break issue seems to be gone 👍 

But unfortunately there is now a little glitch with the copied string:
![Annotation 2019-05-20 102349](https://user-images.githubusercontent.com/11785076/58007192-0832e580-7aea-11e9-8f6a-17d038a9cb4b.png)

Issue #97, opened 2018-10-09
vim interface broken, new line on every keypress


jonaskuske:
When using vi/vim, a new line is created on every keypress instead of overriding the current one(s).
So writing "Hi" looks like this:
```bash
~
H
~
Hi
~
```
The 'static interface' at the bottom (mode, filename, cursor position etc.) is duplicated as well every time, which pushes everything else up and out of the window - so before you scroll, you don't even see what you've written. Here is a GIF showing the issue:

![fluentissue](https://user-images.githubusercontent.com/30421456/46700896-5b70bb80-cc1e-11e8-81d1-365d8f397eab.gif)

Editing using nano works without problems and vim in Hyper also works flawlessly.

I'm using WSL (Ubuntu) on Windows 1809

skipmeister123:
I've experienced this too.
--
Hans Kokx

felixse:
Is this the released version (0.2.0.0) or current master? Also, did it work before updating to 1809?

jonaskuske:
The released version.
Can't say if it worked on 1803, all my devices are on 1809 and I first installed FluentTerminal after the feature update.

felixse:
This seems be fixed with the latest master. Probably the WinPTY update from #95

Riebart:
I have had this crop up on me from time to time outside of vim, but not in any reproducible way. A new tab in the same window does not experience this issue when it rears its head in one of my tabs. I can confirm it's a thing, but can't offer any details to point to a cause.

ad-on-is:
Same here, just had this problem outside of vim. 
After restarting FluentTerminal the problem persisted in the starting tab, while opening a new tab, did not have this kind of issue.

Prior to that I noticed that FT used 900MB of memory, which may be caused by the large log-outputs I'm monitoring during web-development with Docker. I've set the scrollback limit to 999999.

Maybe this helps.

![image](https://user-images.githubusercontent.com/9553253/47212024-5edf1200-d397-11e8-9d01-309b6ae968dd.png)

StiviiK:
Similar issues when using `nano`

nicohouillon:
in VI, if I want to move my cursor up/down it would print out character : A or B . I f I move right/left , then ir printout C or D .
only when I am in edition mode in VI  , after having pressed "I"or "A" .

nicohouillon:
ok,  `set nocompatible` in ~/.vimrc fixed my problem above, still it was working fine in Windows Terminal Preview without this setting.

PR #172, opened 2019-01-12
Line ending translation

Riebart:
Adds a new property to `ShellProfile`s, as well as the settings pages, that determines how to translate line endings (Convert all to CRLF, Convert all to LF, and Do not modify) on paste. Conversion is done through regular expression replacement.

Also adds a new command: Paste without newlines, which uses the regular expression to replace \r and \n characters with empty strings. Useful for handling pasting copied multi-line commands that may have had a newline introduced into them through soft wrapping.

Addresses #110


Riebart:
@skipmeister123 Maybe? Depends on whether that issue is line-ending related, or related to other bugs where newlines appear after each typed character (#97). Notably, I can no longer reproduce the latter, and this should fix anything related to the former.

It can also be used as a workaround for #169 in some cases.

skipmeister123:
I just pulled down the latest code and tested it.  Looks like it is still broken...

--
Hans Kokx
On 1/12/2019 4:39:16 PM, Mike <notifications@github.com> wrote:
@skipmeister123 [https://github.com/skipmeister123] Maybe? Depends on whether that issue is line-ending related, or related to other bugs where newlines appear after each typed character (#97 [https://github.com/felixse/FluentTerminal/issues/97]). Notably, I can no longer reproduce the latter, and this should fix anything related to the former.
It can also be used as a workaround for #169 [https://github.com/felixse/FluentTerminal/issues/169] in some cases.

Riebart:
I'm without my dev environment for a bit, can you confirm that when you paste into other WSL processes, like nano and bash, that the line endings are coming out correctly.

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
```diff
@@ -238,7 +238,7 @@
                                     Text="{x:Bind Arguments, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" />
                                 <ComboBox
                                     Width="150"
-                                    Margin="0,24,0,0"
+                                    Margin="0,24,0,16"
```
ericcornelissen: This seems to be an accidental/unnecessary/unrelated change?

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
```diff
@@ -264,6 +264,19 @@
                                     <HyperlinkButton Command="{x:Bind AddKeyboardShortcutCommand}" Content="Add" />
                                 </StackPanel>
                                 <views:KeyBindingsView ShowCommandName="False" ViewModel="{x:Bind KeyBindings}" />
+                                <TextBlock Margin="0,0,0,8" Text="Line ending translation on paste" />
```
ericcornelissen: I would expect the margins to be either `0,24,0,8` or `0,36,0,8`

```suggestion
                                <TextBlock Margin="0,36,0,8" Text="Line ending translation on paste" />
```

ericcornelissen: In spirit of the other option titles in the shell profile configuration I would rephrase the text as ⬇️

If you update this, don't forget to update it [here](https://github.com/felixse/FluentTerminal/blob/4349d3ad44169f4de38365a3fc00568f47edbeaa/FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml#L324) as well.

```suggestion
                                <TextBlock Margin="0,0,0,8" Text="Line ending translation" />
```

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
```diff
@@ -264,6 +264,19 @@
...
+                                <RadioButton
+                                    Content="Do not modify line endings on paste"
+                                    GroupName="LineEndingStyle"
+                                    IsChecked="{x:Bind LineEndingStyle, Mode=TwoWay, Converter={StaticResource LineEndingStyleToRadioButtonConverter}, ConverterParameter='DoNotModify'}" />
+                                <RadioButton
+                                    Content="Convert line endings to CR+LF on paste"
+                                    GroupName="LineEndingStyle"
+                                    IsChecked="{x:Bind LineEndingStyle, Mode=TwoWay, Converter={StaticResource LineEndingStyleToRadioButtonConverter}, ConverterParameter='ToCRLF'}" />
+                                <RadioButton
+                                    Content="Convert line endings to LF on paste"
```
ericcornelissen: I would personally go with shorter descriptions here, e.g.: "Do not modify", "Convert to CRLF", and "Convert to LF". Although it is personal preference, you seem to have done the same in the preview of the settings.

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
ericcornelissen: Since the "Keyboard shortcut" setting can grow in size it feels awkward to me that this setting is placed below that. Personally I would move this setting above the "Keyboard shortcut" setting.

If you update this, don't forget to update it [here](https://github.com/felixse/FluentTerminal/blob/4349d3ad44169f4de38365a3fc00568f47edbeaa/FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml#L324) as well.

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
```diff
@@ -307,6 +320,9 @@
                                         ViewModel="{x:Bind KeyBindings}" />
                                 </controls:HeaderedContentControl>
 
+
+                                <TextBlock Margin="0,0,0,8" Text="Line ending translation on paste" />
```
ericcornelissen: I would expect the margins to be either `0,26,0,8`

```suggestion
                                <TextBlock Margin="0,26,0,8" Text="Line ending translation on paste" />
```

ericcornelissen: The way this is written is inconsistent with the other options, I think it should be more like this:

```suggestion
                                <controls:HeaderedContentControl
                                    Margin="0,36,0,0"
                                    HorizontalContentAlignment="Stretch"
                                    Header="Line ending translation on paste">
                                    <ContentControl
                                        Margin="0,6,0,0"
                                        Content="{x:Bind LineEndingStyle, Mode=OneWay, Converter={StaticResource EnumToDescriptionConverter}}"
                                        ContentTemplateSelector="{StaticResource TerminalThemeTemplateSelector}" />
                                </controls:HeaderedContentControl>
```

FluentTerminal.App.ViewModels/TerminalViewModel.cs
```diff
@@ -341,11 +341,22 @@ private async void OnKeyboardCommandReceived(object sender, string command)
...
+                string content = await _clipboardService.GetText().ConfigureAwait(true);
                 if (content != null)
                 {
+                    // Use the shell profile line-ending translation method to translate the line endings of the pasted buffer
+                    // or delete all of the instances of the newline pattern by replacing with the empty string.
+                    if (command == nameof(Command.Paste))
+                    {
+                        content = _shellProfile.TranslateLineEndings(content);
+                    }
+                    else
```
ericcornelissen: Perhaps make more explicit for clarity (don't know if the maintainers are a fan of this though)?

```suggestion
                    else if (command == nameof(Command.PasteWithoutNewlines))
```

FluentTerminal.App.ViewModels/TerminalViewModel.cs
```diff
@@ -341,11 +341,22 @@ private async void OnKeyboardCommandReceived(object sender, string command)
...
             }
-            else if (command == nameof(Command.Paste))
+            else if ((command == nameof(Command.Paste)) || (command == nameof(Command.PasteWithoutNewlines)))
             {
-                var content = await _clipboardService.GetText().ConfigureAwait(true);
+                string content = await _clipboardService.GetText().ConfigureAwait(true);
                 if (content != null)
                 {
+                    // Use the shell profile line-ending translation method to translate the line endings of the pasted buffer
+                    // or delete all of the instances of the newline pattern by replacing with the empty string.
```
ericcornelissen: Maybe inline the comments inside the if-statement? To be honest I don't think they're really even necessary...

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
Riebart: Changed to `Line ending translation mode`

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
Riebart: The idea is to keep descriptions short when viewing a config, but descriptive when editing the configuration, to allow people to understand the functionality when they go to edit a profile. Leaving it as is for now.

FluentTerminal.App.ViewModels/TerminalViewModel.cs
Riebart: Resolved by converting that section to a `switch() { }`.

FluentTerminal.App.ViewModels/TerminalViewModel.cs
Riebart: Resolved by converting that section to a `switch() { }`.

Riebart:
Accepted most of Eric's suggestions. See most recent commit.

FluentTerminal.App.ViewModels/TerminalViewModel.cs
ericcornelissen: Great solution 👍

Stylistically I'm unsure whether the braces in `case: { }` are desired, I've personally never seen it and as far as I can tell it's not used anywhere else in the project. That said, don't feel the need to change it unless the maintainer(s) say something about it 😄

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
```diff
@@ -280,7 +293,7 @@
                                     Header="Arguments"
                                     Text="{x:Bind Arguments, Mode=OneWay}" />
                                 <controls:HeaderedContentControl
-                                    Margin="0,36,0,0"
+                                    Margin="0,36,0,48"
```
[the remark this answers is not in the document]
ericcornelissen: Can you revert this change? It seems to be causing this gap ⬇️

gap preview

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
ericcornelissen: Can you look at this change again? It seems to be causing this gap ⬇️

gap preview

FluentTerminal.App/Views/SettingsPages/ShellProfileSettings.xaml
ericcornelissen: Hmm, that is a good motivation 🤔 Thinking back, I think my main motivation for commenting was that you're repeating "line endings ... on paste" unnecessarily, what about changing it to something like ⬇️?

* * *

Line endings translation on paste
- [ ] Do not modify
- [ ] Convert to CR+LF
- [ ] Convert to LF

* * *

Or in case you want to keep "line endings" for clarity ⬇️:

* * *

Line endings translation on paste
- [ ] Do not modify line endings
- [ ] Convert line endings to CR+LF
- [ ] Convert line endings to LF

FluentTerminal.App.ViewModels/TerminalViewModel.cs
Riebart: Stylistically, that's a hold over from ANSI style bracing of C++, and is (at least, in C++), required if you ever define a local scope variable inside a switch case like we do for the two Paste commands). I'm not married to it, but trained myself over years of using `astyle -j --style ansi`.

ericcornelissen: Makes sense, didn't know that. Thanks for the explanation 🙂

Riebart:
Aside from the typo in the commit message, we should be good to merge I think.

felixse:
Great work 👍 
I pushed some changes, mostly about the way the radio button binding is implemented. I know the way I implemented it is a bit verbose, but I had my fair share of troubles using the converter approach in UWP. Plan is to wrap this in a nice little control in the future to keep the ViewModels clean.

PR #173, opened 2019-01-15
Better menu

skipmeister123:
Replaced the hamburger dropdown menu with a slick sliding panel

image

### DIFF
--- a/FluentTerminal.App.Services.Test/FluentTerminal.App.Services.Test.csproj
+++ b/FluentTerminal.App.Services.Test/FluentTerminal.App.Services.Test.csproj
@@ -11,13 +11,16 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="AutoFixture" Version="4.4.0" />
-    <PackageReference Include="FluentAssertions" Version="5.4.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.0" />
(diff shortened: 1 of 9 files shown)


skipmeister123:
To be fair, I have no idea what I'm doing - I don't know C# or XAML. I just 
wanted a proof of concept to open a conversation and let somebody who knows 
something fix it 😅
--
Hans Kokx

skipmeister123:
Oh, also, I couldn't figure out how to have the side bar close automatically once you click an item.  Seems like something that would need to be done in C#, which I spent a day trying to figure out and gave up on.  I'm also thinking about an option to have profiles listed under the new tab button -- so you'd have "new tab", "new powershell tab", "new cmd tab", "new wsl tab", etc.  Each profile could have its own icon, defined in the settings.  Maybe we could use http://nerdfonts.com/ to provide icons?

skipmeister123:
Oh, one last thing: I wanted the side panel to be Acrylic, like the side of the settings window... but I failed to figure that out.

skipmeister123:
@felixse Can you weigh in on this PR?

felixse:
Yes sure. I like the idea as well, just a bit short on time right now. Plan is to complete the Conpty option first, after that I should be in the mood for some more visual changes 😄

felixse:
I will close this one in favor of the other better menu PR

PR #183, opened 2019-01-20
Add option to use bold text in terminals

### LINKED ISSUE #182: Allow bold text for prompts etc
ConsoleZ (right hand side) has an option to enable bold text for prompts etc (which is how the native `gnome-terminal` looks). Is this possible (or could it be implemented :smile:)?

screenshot_win10_2019-01-20_16 17 50

ericcornelissen:
Closes #182

Adds a toggle to the Terminal Page to enable/disable bold text in terminals. Currently, from my testing, when `BoldText` is enabled there is no real distinction between normal and bold text in most fonts, I also addressed this in [this comment](https://github.com/felixse/FluentTerminal/issues/182#issuecomment-455901102).

### Alternative implementation
Currently the conversion from boolean to `fontWeight` and `fontWeightBold` value is performed 2x each, alternatively we could use transformer functions, e.g. ⬇️, but this seems like overkill to me.

```javascript
term.setOption('fontWeight', boldToFontWeight(options.boldText));
term.setOption('fontWeightBold', boldToFontWeightBold(options.boldText));

...

function boldToFontWeight(boldText) {
  return options.boldText ? 'bold' : 'normal';
}
function boldToFontWeightBold(boldText) {
  return options.boldText ? 'bolder' : 'bold';
}
```

### DIFF
--- a/FluentTerminal.App.ViewModels/Settings/TerminalPageViewModel.cs
+++ b/FluentTerminal.App.ViewModels/Settings/TerminalPageViewModel.cs
@@ -136,6 +136,20 @@ namespace FluentTerminal.App.ViewModels.Settings
             }
         }
 
+        public bool BoldText
+        {
+            get => _terminalOptions.BoldText;
+            set
+            {
+                if (_terminalOptions.BoldText != value)
+                {
+                    _terminalOptions.BoldText = value;
+                    _settingsService.SaveTerminalOptions(_terminalOptions);
+                    this.RaisePropertyChanged();
+                }
+            }
+        }
+
         public IEnumerable<string> Fonts { get; }
 
         public int FontSize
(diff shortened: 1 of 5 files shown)


FluentTerminal.App.ViewModels/Settings/TerminalPageViewModel.cs
```diff
@@ -136,6 +136,20 @@ public string ScrollBackLimit
             }
         }
 
+        public bool BoldText
```
felixse: Please also add to TerminalPageViewModel.RestoreDefaults()

FluentTerminal.Models/TerminalOptions.cs
```diff
@@ -8,6 +8,8 @@ public class TerminalOptions
 
         public int FontSize { get; set; }
 
+        public bool BoldText { get; set; }
```
felixse: Can you please specify a default value for this in DefaultValueProvider?

ericcornelissen:
The build failure seems to be caused by [the certificate being invalid](https://ci.appveyor.com/project/felixse/fluentterminal/builds/21909961#L826), the same is true for the [latest build on master](https://ci.appveyor.com/project/felixse/fluentterminal/builds/21906620#L825) even though the certificate was updated in that commit for #185 🤔

felixse:
That's my fault. The CI is using some encrypted certificates I should also have updated, but my plan is to migrate the CI to Azure Pipelines in the next days.

PR #194, opened 2019-02-05
Suggestion: Better CLI parsing, command

fabian0010:
I added the option to run a command via the CLI.
Also added a better argument parsing solution.

You can now do `flute.exe new "[PATH]" "[COMMAND]"`
Before, you could only do `flute.exe new [PATH]`

You would also need to adjust your settings, say your default profile is using powershell with custom arguments, those arguments would have to end with ";", because the `[COMMAND]` would be appended to the arguments.

image
image

I went ahead and made those changes for my own personal needs, but I think others could benefit from this as well


fabian0010:
> What are some use cases for this? -- Hans Kokx


I'm personally want to use FluentTerminal as en external debugging terminal for vscode, which i would need that for

skipmeister123:
I like where you're going with this.  I haven't reviewed the code, nor tried it out, but if the work is done we should pull it in. 👍


--
Hans Kokx
On 2/5/2019 12:28:37 PM, Fabian <notifications@github.com> wrote:
In general, I thought it was a good idea to change the argument parsing since now it would allow for more advanced CLI commands. It would also be great to enable the user to run a command from the CLI, just like all those other terminals do, for example cmd /k [COMMAND], powershell [COMMAND], terminus run [COMMAND]. This would be used for, as I said, starting FluentTerminal with a debugger.
But it's just a suggestion, I really like this project and I'd love to contribute ^^

felixse:
I really like the idea, but I think we might run into problems if we keep adding arguments without a real structure. We could add something like [commandline](https://github.com/commandlineparser/commandline) to make our CLI verb based and more flexibel.

felixse:
Ok, I integrated https://github.com/commandlineparser/commandline
There are 2 verbs for now:
* New:
accepts a directory as a anonymous optional parameter
accepts a command, e.g. --command "your command"
(we could also add something to select the profile)

* Verb:
no options for now, could be extended to open a specific page

Next step would be to refactor the current AddTerminal() to accept some LaunchArguments class instead of this growing number of options

felixse:
Ok, running commands is working fine now. Since commands are handled a bit different than profiles I think we should give them their own verb.

My current proposal would be something like this:

```
flute new "C:\" --profile "wsl"  --target=tab|window
flute run "npm start" --directory "C:\" --theme="ubuntu bash" --target=tab|window
flute settings general|terminal|etc...
```

Any suggestions or ideas?

ericcornelissen:
Primary suggestion I have is to add a **help** command/verb as well.

That said, the `--target` is a bit buggy for me. When using `window` the new window (but that might be due to #206?). When using `tab` it sometimes closes some or all of the preexisting tabs.

Also, I'm not sure what you exactly want the run command to do but I would expect the new terminal to give me some output. Also, when I run a command like `ls` it immediately closes the terminal (not completely unexpected, but maybe add a `--keep-alive` option?), but when I run a command like `npm test` the terminal seems to stay open forever (at least way longer than I'm expecting).

Lastly, though I'm not sure if you finished this yet, the `settings` verb seems to always open the settings in General.

felixse:
> Primary suggestion I have is to add a **help** command/verb as well.
> 

The problem is that we cannot write a response using the appExecutionAlias😞 

> That said, the `--target` is a bit buggy for me. When using `window` the new window (but that might be due to #206?). When using `tab` it sometimes closes some or all of the preexisting tabs.
> 

Anything you can reproduce? Works good for me

> Also, I'm not sure what you exactly want the run command to do but I would expect the new terminal to give me some output. Also, when I run a command like `ls` it immediately closes the terminal (not completely unexpected, but maybe add a `--keep-alive` option?), but when I run a command like `npm test` the terminal seems to stay open forever (at least way longer than I'm expecting).
> 

That's because there is no shell process to return to. From what I understand this behavior is the same as in other terminals.

> Lastly, though I'm not sure if you finished this yet, the `settings` verb seems to always open the settings in General.

I think I will skip this, not too much value to gain here 😅

ericcornelissen:
> The problem is that we cannot write a response using the appExecutionAlias

😞 


> I think I will skip this, not too much value to gain here 😅

Sure, just thought I would point it out. Maybe convert it into an issue labeled `Good first issue`?


> That's because there is no shell process to return to. From what I understand this behavior is the same as in other terminals.

I'm assuming your response is with regards to the output? What about the inconsistent behavior of automatically closing vs. staying open for different commands?

To be honest, I've never used functionality like this before so I don't know what to expect.


> Anything you can reproduce? Works good for me

This seems to work for the `--target=tab` case (hope my ad hoc notation is clear):

```bash
{W1,T1} $ cd D:\foo
{W1,T1} $ flute new "D:\bar\" --profile "powershell"  --target=tab
{W1,T2} $ exit
{W1,T1} $ flute new "D:\bar\" --profile "powershell"  --target=window
{W2,T1} $ exit
{W1,T1} $ flute new "D:\bar\" --profile "powershell"  --target=tab
# {W1,T1} is gone and only {W1,T3} is present
```

felixse:
> > I think I will skip this, not too much value to gain here 😅
> 
> Sure, just thought I would point it out. Maybe convert it into an issue labeled `Good first issue`?

Good idea 👍 
> > That's because there is no shell process to return to. From what I understand this behavior is the same as in other terminals.
> 
> I'm assuming your response is with regards to the output? What about the inconsistent behavior of automatically closing vs. staying open for different commands?
> 
> To be honest, I've never used functionality like this before so I don't know what to expect.
> 

Me neither. ls is behaving as I would expect it. I don't know what is going on with npm test though.

> > Anything you can reproduce? Works good for me
> 
> This seems to work for the `--target=tab` case (hope my ad hoc notation is clear):
> 
> ```shell
> {W1,T1} $ cd D:\foo
> {W1,T1} $ flute new "D:\bar\" --profile "powershell"  --target=tab
> {W1,T2} $ exit
> {W1,T1} $ flute new "D:\bar\" --profile "powershell"  --target=window
> {W2,T1} $ exit
> {W1,T1} $ flute new "D:\bar\" --profile "powershell"  --target=tab
> # {W1,T1} is gone and only {W1,T3} is present
> ```

The 'exit closes a window' scenario was a leftover of #206. Should be fixed now. Thanks for finding this 😃

ericcornelissen:
> Me neither. ls is behaving as I would expect it. I don't know what is going on with npm test though.

In that case you can leave it for now I guess. There will be bug reports from people expecting different behavior sooner or later.


> The 'exit closes a window' scenario was a leftover of #206. Should be fixed now. 

Seems to be fixed, thanks 👌 

* * *

I just noticed that for some reason the PowerShell message is showing on this branch. I'm using the `-nologo` argument in the PowerShell profile, so I'm guessing that is not being picked up?

```
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.


Loading personal and system profiles took 1049ms.
erico@LAPTOP-ERIC D:\foo
```

felixse:
Just had a try with npm test. I guess you just saw an empty window? That happens when it failed to start. I'll see if I can optimize that a bit further. you need to call 'npm.cmd' and pass the correct directory. Another todo here is to pass the cwd of the caller if none is specified. But then it launched correctly and stayed open until I quitted it.

I'll have a look at the arguments. Might explain the strange stuff I encountered in WinPtySession 😄

ericcornelissen:
> Just had a try with npm test. I guess you just saw an empty window? That happens when it failed to start. I'll see if I can optimize that a bit further. you need to call 'npm.cmd' and pass the correct directory. Another todo here is to pass the cwd of the caller if none is specified. But then it launched correctly and stayed open until I quitted it.

Hmm 🤔 would be nice if it just worked with `npm` rather then `npm.cmd`...


> I'll have a look at the arguments. Might explain the strange stuff I encountered in WinPtySession 😄

Seems to be fixed 👍

PR #196, opened 2019-02-05
Better Menu Fix

fabian0010:
I fixed #173 

image


fabian0010:
Okay after a bunch of trial and error i realized that the original better-menu branch is outdated, it should theoretically all be good now.

Todo: find out why the menu isn't animated (at least for me)

skipmeister123:
SORRY. 😭
--
Hans Kokx

skipmeister123:
Also... It's animated for me on 1809? And check the framework updates from 
my original PR... I think that has something to do with it.
--
Hans Kokx

felixse:
I think we should avoid to updated the target framework when possible. My plan for the menu is to actually use a [NavigationView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview) in minimal mode, this gives us a very system like design by default, also animations.

fabian0010:
@skipmeister123  haha it's fine ^^

@felixse yeah i thought so as well, so this branch runs on the original target framework now

oh, I didn't know about the minimal mode, I gotta check that out

fabian0010:
Oh my, that minimal mode is perfect, why didn't I find this anywhere

fabian0010:
Nevermind, to use it we'd need a greater target version

skipmeister123:
> 
> 
> I think we should avoid to updated the target framework when possible. My plan for the menu is to actually use a [NavigationView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview) in minimal mode, this gives us a very system like design by default, also animations.

I wanted to avoid that, because I think the back arrow is ugly and unnecessary.

patriksvensson:
> I wanted to avoid that, because I think the back arrow is ugly and unnecessary.

@skipmeister123 The back arrow is configurable. 
A NavigationView sounds like a perfect fit for what you're trying to accomplish.

fabian0010:
@skipmeister123 you can deactivate the back arrow, that wouldn't be a problem, what would be a problem is that the NavigationView only seems to support said minimal mode (which only shows the button when the pane is closed) in a more recent target version

felixse:
There is no PaneDisplayMode, but we should be able to achieve this by setting both threshholds to something very big

fabian0010:
No clue what you mean, but sounds good ^^

skipmeister123:
Hey, we could pull some Microsoft logic for handling the menu.  I like how Calculator does it, anyway.
https://github.com/Microsoft/calculator/tree/master/src/Calculator

felixse:
Didn't know that calculator is open source, thanks for posting this 👍 
So far it looks like NavigationView is not the right thing to use. There is a lot of workaround to do because we don't navigate but try show a menu instead and not having the button somewhere over in the terminal. I think it will be easier to follow the SplitView approach with DisplayMode set to Overlay

felixse:
Currently investigating in this direction:
image

Todo:
* Add icons
* Remove height
* Add keybinding information

What do you think?

skipmeister123:
Maybe add "FluentTerminal" to the bottom of the menu? It looks pretty naked 
right now. How does it handle vertical resizing? Is it a scroll box?
--
Hans Kokx

ericcornelissen:
Looks awesome! I like the idea of adding the keybinding info, any idea on how that will look? Maybe instead of the lines you can have some sort of header (not sure what the text of the header would be though...)?

> Maybe add "FluentTerminal" to the bottom of the menu? It looks pretty naked right now.

Alternatively you could move the "About" button there similar to the settings window and the Calculate app. Or perhaps "Quit"?

skipmeister123:
> Alternatively you could move the "About" button there similar to the settings window and the Calculate app. Or perhaps "Quit"? Maybe both?

Kinda like I did before? :P

felixse:
We could also add the content of the about page there. That way we have the app name, the current version, and maybe even an update indicator there

ericcornelissen:
> Kinda like I did before? :P

Hehe 😅 Apparently, yes


> We could also add the content of the about page there. That way we have the app name, the current version, and maybe even an update indicator there

Not a bad idea, the main argument against that I think is that it is not consistent with other UWP apps (at least in my experience). Also, it might get a little bit cluttered that way 🤔

felixse:
Update:
image

skipmeister123:
I love it, but I'd like the ℹ️ icon instead of the ?

felixse:
> I love it, but I'd like the ℹ️ icon instead of the ?

Yes of course, I was just hacking this together and was to lazy to search for the right icon.

The bad news is that I will need to bump the TargetVersion to 1803 for this. I think we should release 0.4.0.0 first, then raise the version for 0.5.0.0

felixse:
I thought about giving 'New Tab' and  'New Window' submenus with elements for each profile. Just wondering how to add the default. Maybe if you click the root menu item?

0.4.0.0 should be coming this week The plan is to merge the command PR, add the TERM_PROGRAM environment variable and some logging improvements and then release it.

ericcornelissen:
> I thought about giving 'New Tab' and 'New Window' submenus with elements for each profile. Just wondering how to add the default. Maybe if you click the root menu item?

I was thinking something like [PaneCustomContent](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#pane-content) (although I would put the arrow at the end rather than the start in this particular use case). I'm assuming the `Item 3a` only shows after you clicked `item 3`.

preview from UWP docs


Or perhaps a right click/context menu? But I don't think that is a good choice from a usability stand-point (its not obvious).


> 0.4.0.0 should be coming this week The plan is to merge the command PR, add the TERM_PROGRAM environment variable and some logging improvements and then release it.

Ah oke, in that case its definitely better to wait 👍

NikkyAI:
> I thought about giving 'New Tab' and 'New Window' submenus with elements for each profile. Just wondering how to add the default. Maybe if you click the root menu item?

how about having it read `+` `Powershell` as 2 buttons? blicking the `+` would open the dropdown with all non-default profiles

in gitkraken: 

this works and is very intuitive, in these you can also switch the defaults by clicking on the radio buttons
i feel like quickly switching the defaults based on what kind of task i am doing might be useful, but there could also be misclicking, in that case maybe require holding a modifier key to switch defaults..
anyways the othjer mockups there also look great

felixse:
image

How about this?

skipmeister123:
It's better, but not scalable. If you have a lot of bookmarked remote 
servers or profiles, the list will be unwieldily.  Having the sidebar would 
allow you to scroll.

felixse:
What are the chances people have more than 10 profiles? The sidebar just looks a bit off, and I don't think we should optimize for the edge case

skipmeister123:
I have a lot of lab boxes which I'd create profiles for. I would easily 
exceed 10.

PR #199, opened 2019-02-08
Experimental ConPTY support

felixse:
Can be enabled in Settings -> General

Known issues:
- Tab title might include the full path of the shell executable
- PowerShell: Cursor is flickering in first row, second column on input
- Color 243 is a wrong grey tone
- Random weirdness


hchokshi:
I'm currently trying to put together a fix for the tab title issue. Hopefully have a PR some time in the next few days :)

skipmeister123:
Thank you!

hchokshi:
It looks like that full path is coming from an escape sequence (either `OSC 0 - set icon and title` or `OSC 2 - set title`) being sent by `cmd`/`powershell`/`wsl` with their full path as the title.

Doesn't seem like a good idea to special case these paths, so not sure about the best way to achieve desired behaviour without potentially affecting other programs trying to set the title.

felixse:
So the problem is that WinPty doesn't do this. We should modify the part where the fallback title for WinPty is set to be in line with ConPty

PR #204, opened 2019-02-15
installation switch to force context menu integration automatically

pgalbraith:
Allows scripted install of context menu integration with `-ForceContextMenu` switch


felixse:
Thanks 👍

PR #213, opened 2019-02-20
Improve settings window

ericcornelissen:
I tried to improve the appearance and usability of the settings window mainly by using consistent margins, and add some quality of life improvements.

- [x] Normalized margins between settings
- [x] Always have some margin at the bottom of a page
- [x] Use constant width for text inputs
- [x] Added default display values for a Profile's "Working directory", "Arguments", and "Keyboard shortcut"
- [x] Attempt to make `.xaml` files more readable
- [x] When cancelling creating a new (or cloning) a profile, delete the new profile
- [x] Open "Create new theme" in edit mode
- [x] Don't prompt user for cancelling without changing anything in the Profile
- [x] Don't prompt user for cancelling without changing anything in the Themes
- [x] Update "loading..." text in About page


FluentTerminal.App/Views/SettingsPages/About.xaml
```diff
@@ -52,36 +50,40 @@
...
-                    <TextBlock Visibility="{x:Bind ViewModel.LatestVersionFound, Mode=OneWay, Converter={StaticResource TrueToVisibleConverter}}">
-                        <Run Text="{x:Bind ViewModel.LatestVersion, Mode=OneWay}" />
-                         - <Hyperlink NavigateUri="{x:Bind ViewModel.LatestVersionReleaseNotesURL, Mode=OneWay}">
-                            Release notes
-                        </Hyperlink>
-                    </TextBlock>
+                    <StackPanel Margin="0,24,0,0">
+                        <TextBlock Margin="0,0,0,8" Text="Latest version" />
+                        <TextBlock Visibility="{x:Bind ViewModel.LatestVersionLoading, Mode=OneWay, Converter={StaticResource TrueToVisibleConverter}}">
+                            <Italic>loading...</Italic>
```
skipmeister123: Should this be capitalized? I feel like maybe "Checking for update..." would work better? Or maybe a spinner? I dunno. This just doesn't seem quite right.

ericcornelissen: That's a good a suggestion, I'll definitely update the text and I'll see about a spinner

ericcornelissen:
What I did so far was manually update the margins of all elements in the `.xaml` files. Alternatively we could move this to a [XAML style](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/xaml-styles) which would have two major benefits:

1. Easier to change the overall margins
1. Contributors are less likely to make a mistake regarding margins in the future

Thoughts @felixse?

skipmeister123:
Having this as a single variable seems like the best idea.

felixse:
Good changes here, thank you 👍 

> Having this as a single variable seems like the best idea.


Yes, that actually how it should be done, just hacked this together initially and never bothered cleaning it up.

I have introduced two margin definitions:
ContainerMargin is defining margin left, right and bottom and should be used for the element that contains that items that should have equal margin.
ItemMargin is defining margin at the bottom, just for all direct children of the container.

Besides that I had some issues with the new Equals method in terminal theme, we should consider that author could also be null. Was running out of time tonight, but I can have a look at this tomorrow

FluentTerminal.App/Views/SettingsPages/About.xaml
```diff
@@ -45,43 +45,46 @@
...
-                <StackPanel Grid.Row="1" Margin="0,72,0,24" VerticalAlignment="Bottom">
+                <StackPanel
+                    Grid.Row="1"
+                    Margin="0,72,0,0"
+                    VerticalAlignment="Bottom">
                     <TextBlock>
-                        Created by <Italic>felixse</Italic>
-                        and <Hyperlink NavigateUri="https://github.com/felixse/FluentTerminal/graphs/contributors">contributors</Hyperlink>
+                        Created by<Italic>felixse</Italic>
+                        and<Hyperlink NavigateUri="https://github.com/felixse/FluentTerminal/graphs/contributors">contributors</Hyperlink>
```
ericcornelissen: Don't think its intentional, but it seems you removed some necessary whitespace in this file by accident/automatically

FluentTerminal.App/Views/SettingsPages/About.xaml
```diff
@@ -12,7 +12,7 @@
             <RowDefinition Height="Auto" />
             <RowDefinition Height="*" />
         </Grid.RowDefinitions>
-        <Grid Grid.Row="1" Margin="24,0,0,24">
+        <Grid Grid.Row="1" Margin="24,0,0,0">
```
ericcornelissen: Not as important, but we might consider defining a `<Thickness>` for this margin as well

ericcornelissen:
> I have introduced two margin definitions [...]

I added a third one for the header margins just for completeness.


> Besides that I had some issues with the new Equals method in terminal theme, we should consider that author could also be null. Was running out of time tonight, but I can have a look at this tomorrow

I fixed this particular instance where a variable can be `null`. Alternatively, I think, we can instead use `Object.equals(obj1, obj2)` (see [docs](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=netframework-4.7.2#System_Object_Equals_System_Object_System_Object_)) everywhere, but I'm not 100% that will work as intended, so I didn't update it yet...

FluentTerminal.App/Views/SettingsPages/About.xaml
felixse: Damnit, not again. Xaml Styler is quite aggressive on whitespace. I will move the text to the code behind

PR #193, opened 2019-02-03
Add keybinding to change the (current) tab title

ericcornelissen:
As per [the suggestion](https://github.com/felixse/FluentTerminal/issues/174#issuecomment-454673472) by @ChristianSauer. I set the default binding to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd>.

Issue #96, opened 2018-10-09
Add keybinding for "Configurable new window"


jonaskuske:
There already is the option to open a "configurable new tab", which opens a new tab with a profile picker, letting you select one of your profiles before loading the shell.
A similar option for windows, opening a new window and letting you select a profile first would be nice to have. (`Ctrl+Shift+N` by default?)

felixse:
Should be no problem

PR #216, opened 2019-02-27
Add keybinding for configurable new window

ericcornelissen:
As per #96 and [this comment](https://github.com/felixse/FluentTerminal/pull/193#issuecomment-460706953) I added a keybinding to open a new configurable window. The default binding is <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd>.

preview

### Alternative design

1] It may be possible to rely less on a parameter (`showProfileSelection`) in _App.xaml.cs_, but I personally don't think that's a good idea.

* * *

2] There are now two `EventArgs` (namely `CancelableEventArgs` and `ProfileSelectEventArgs`) that basically only hold a boolean value (`cancelled` and `ShowProfileSelection` respectively). On the one hand one the naming helps the reader understand the code, on the other hand it you have (albeit simple) duplicate code. I can refactor the `EventArgs` into a single `BooleanEventArgs` class like so:

```c#
// Current use of CancelableEventArgs
e.Cancelled = true;

// Example of using BooleanEventArgs
cancelable.Value = true;
```

However, the current approach forces the programmer to be descriptive, whereas the refactor would allow misuse of the `EventArgs`, like so:
```#
// Misuse of BooleanEventArgs
e.Value = true;
```

### DIFF
--- a/FluentTerminal.App.Services/EventArgs/NewWindowRequestedEventArgs.cs
+++ b/FluentTerminal.App.Services/EventArgs/NewWindowRequestedEventArgs.cs
@@ -1,0 +1,7 @@
+﻿namespace FluentTerminal.App.Services.EventArgs
+{
+    public class NewWindowRequestedEventArgs : System.EventArgs
+    {
+        public bool ShowProfileSelection { get; set; }
+    }
+}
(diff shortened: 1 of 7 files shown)


felixse:
great job 👍 

I think a little duplication here is fine, it makes the code more clear this way. I also renamed the new EventArgs to describe the context in which it is used.

ericcornelissen:
> I think a little duplication here is fine, it makes the code more clear this way

Agreed

> I also renamed the new EventArgs to describe the context in which it is used.

I wasn't sure about the naming anyway, but this is much better :ok_hand:

PR #217, opened 2019-03-02
[WIP] UWP based renderer

felixse:
I started some investigation into a true UWP renderer. xterm.js is a great library and it really helped us to get where we are now, but I think it is time to invest into a custom renderer to leave the limitations behind we have right now due to having our view basically in another process.

As as first step I isolated everything regarding xterm.js and moved it into XtermTerminalView. This is basically a bigger refactoring of TerminalViewModel and TerminalView. 

The next step would be to create a new implementation based on the UWP Canvas control. Since I want to avoid writing a VT parser from scratch I will make use of [VtNetCore](https://www.nuget.org/packages/VtNetCore/). This project died somewhere in the mid of development so we will most probably need to fork and extend it at some point.


felixse:
VtNetCore doesn't look very promising, guess there would be a lot to do.
My next approach is to wait until https://github.com/xtermjs/xterm.js/issues/1920 is done and then try to port xterm.js to .NET 💀 

I will create a new PR for the encapsulation I did in this branch.

skipmeister123:
What was so unusable about the project, just out of curiosity?
--
Hans Kokx

felixse:
I had a lot of issues with wiggling text and the viewport not acting reasonable. And a miriad of todos in the base library made it pretty hard to distinguish whether it was my fault or something that wasn't implemented completely.

jerch:
> My next approach is to wait until xtermjs/xterm.js#1920 is done and then try to port xterm.js to .NET

That would be doable but quite a big effort. It might be easier using a  C++/C# lib right from the start. Not sure if you could benefit from the new windows terminal here. Others are floating around as well, this lib looks very promising https://github.com/ismail-yilmaz/upp-components/tree/master/CtrlLib/Terminal. Have not looked at it in detail, not sure if its NET compatible.

Issue #141, opened 2018-11-23
Rename tabs


subhead:
Hi,

is it possible to add a feature for renaming tabs? At the moment i need to place my mouse over the tab and read the hover text to get clue ion which terminal i work at the moment.

skipmeister123:
That can be done via environment variables.
--
Hans Kokx

subhead:
yes i know but i find it more convenient to rename the tabs over the gui.

rafd123:
FWIW, as commented on #154,  here's how I do it: 

**cmd.exe**
```
title foo
```

**PowerShell**
```
[console]::Title = 'foo'
```

**bash/WSL**
```
echo -e '\033]2;foo\007'
```

Not suggesting this enhancement shouldn't be implemented, just putting this here as a tip since:
- It tends to work for all terminals (even cmd.exe)
- When wrapped in a shell function, renaming tabs super convenient since you don't have to lift your hands from the keyboard

ericcornelissen:
It seems this issue was resolved with d4aadd7 @felixse?

However, I'm having an issue with the renaming where `npm` (and npm-based commands such as `gulp`) overrides the name and restores the default. For example when I change the title of a tab to "Foo", running `npm install` changes the title to `npm`, and when `npm install` finished the name is restored to "powershell" or "cmd".

PR #220, opened 2019-03-05
Dev/renderer neutral

felixse:
- Improved architecture to support different renderers
- both input and output buffer is now kept in utf8 for the most part, less conversions that might delay
- possibly fixed a race condition, could be #206 and similar

Issue #214, opened 2019-02-21
Startup Default Tabs


SpencerLynn:
I would really love to have a list of tabs that are automatically created at application startup - both the name of the tab and the directory in which the tab should be open.

PR #225, opened 2019-03-08
Persist custom tab title

ericcornelissen:
As I described [in this comment](https://github.com/felixse/FluentTerminal/issues/141#issuecomment-454316202) some processes would overwrite the custom tab title set by the user. Should be fixed with these small changes 😃 

Implementing and testing this made me think that we may consider adding a feature to unset the custom title? Not sure how useful that would be though 🤔


skipmeister123:
We might also want to consider setting it in a custom profile.  For example, if I open a profile that automatically ssh's to a host, I might want that hostname in the tab title.

ericcornelissen:
I like that idea @skipmeister123 (I remember seeing it before 🤔), it's related to #214 as well I guess

felixse:
I think we should also provide a way to go back to shell controlled titles. Right now if you don't put in anything you see no title at all. We should handle empty strings by setting Title to Terminal.FallbackTitle and set _customTitlte to false.

skipmeister123:
I'm going to try to articulate a less than fully formed thought, so bear with me.

I would prefer to set my *tab* title with `precmd() {print -Pn "\e]0;%(4~|%-1~/…/%2~|%4~)\a"}`... but I would also like to have my *window* title be the full command that's running. (I have my tabs on the bottom for this exact reason.)  Currently, there's no way to do that, as the window title is always equal to the tab title.  I recall having the ability to do this in iTerm2.

So, that'd be a feature request.

But building on that thought, if the *window* title and the *tab* title were separate entities, if the tab title was empty, you could simply use a truncated version of the window title.  Make sense?

ericcornelissen:
> I think we should also provide a way to go back to shell controlled titles. Right now if you don't put in anything you see no title at all. We should handle empty strings by setting Title to Terminal.FallbackTitle and set _customTitle to false.

In light of @skipmeister123's comment I did this a little bit different, but when you set the title to `""` it should now go back to the default behaviour.<sup>1</sup>


> I would prefer to set my *tab* title with `precmd() {print -Pn "\e]0;%(4~|%-1~/…/%2~|%4~)\a"}`... but I would also like to have my *window* title be the full command that's running. (I have my tabs on the bottom for this exact reason.)  Currently, there's no way to do that, as the window title is always equal to the tab title.  I recall having the ability to do this in iTerm2.

So basically what you want is that the application title is/can be different from the tab title? I tried to add something like that to this PR. However, I think this should be configurable in some sense, here are the different options I can think of:

- Sync the application and tab title (regardless of whether its edited or not)
- Let the application title be the shell title and have an optional custom tab title
- Have the custom title be overridden when a command is running but restore it when the command is completed

Thoughts? I'm willing to implement this (including configuration) but I think it is out of scope for this PR.

* * *

<sup>1: I tested it from before #220 and rebased to `felixse/master` because of [this](https://github.com/felixse/FluentTerminal/pull/220#issuecomment-471034766), so I'm not 100% sure it is actually working 😅</sup>

skipmeister123:
Here's an example of what I mean.
Screen Shot 2019-03-09 at 10 19 57 AM

Note that the tab text differs from the application title text.

ericcornelissen:
Okay @skipmeister123, then I understood correctly and that should be the current behaviour. But I still think it should be configurable.

felixse:
Fixed some minor issue where ApplicationView.Title (displayed in Windows Taskbar) was not in line with the tab title.

From what I understand we would need to to the following to make @skipmeister123  happy:
* Introduce third title to TerminalViewModel: WindowTitle.
* Bind Titlebar on MainPage to WindowTitle instead of TabTitle
* Add an option whether WindowTitle should be synchronized with TabTitle or ShellTitle
* Bonus: Disable 'Edit Title' in TitleBar (not TabBar) when WindowTitle is set to ShellTitle

ericcornelissen:
> Fixed some minor issue where ApplicationView.Title (displayed in Windows Taskbar) was not in line with the tab title.

Ah, I figured the ApplicationView.Title was the WindowTitle


> Add an option whether WindowTitle should be synchronized with TabTitle or ShellTitle

Does that also mean that ApplicationView.Title is synced with TabTitle or ShellTitle?

Other than that that seems right to me 👍 For me, if I were to adopt the layout of @skipmeister123 (which I'm considering), I would add to that an option to show the command that is running if the WindowTitle is synced with the TabTitle, which I think should be possible using Terminal.FallbackTitle.

danilocode:
There is an issue where the title reverts back to default right after I enter any command even after I have set it to be 'xyz'

ericcornelissen:
@danilocode are you talking about the ApplicationView.Title (what you see in the Windows taskbar) or the tab title in FluentTerminal itself. And in case of the latter, do you have a single tab or multiple tabs. Also, do you have "Always show tabs" enabled?

felixse:
@danilocode Are you using the released version or did you build from current master?

danilocode:
Thanks for the prompt reply, I was out for a few weeks and could not follow up
@ericcornelissen I was talking about tab title. And yes I have multiple tabs opened and yes, always show tabs is also enabled.

@felixse I am using released version

besides great job guys, I really like using this terminal. And I really hope UWP can support administrative tasks as well in future

ericcornelissen:
> Thanks for the prompt reply, I was out for a few weeks and could not follow up

Are you currently running the latest version (`0.4.1.0`) and do you still have the issue (I think your original comment was before the feature in this PR was release)? If so, what command/program are you trying to execute?

PR #237, opened 2019-03-21
Add option to always show original shell title in window title

### LINKED ISSUE #229: Add option: Window title should always show executed command
As discussed in #225

felixse:
Fixes #229 and makes @skipmeister123  happy


ericcornelissen:
I'm not a native English speaker either, but I feel like "Show custom tab title also in titlebar" is not grammatically correct. Could you be persuaded to change it to "Show custom tab title in title bar as well".

* * *

Also, I'm not quite sure about the positioning and wording of the option in general. First of all the option is only applicable if the _Show tabs on_ setting is set to _Bottom_. Second, I'm not sure if the following terms are clear for newcomers: "custom tab title" (might not yet be aware they can change the title) and "title bar" (term is not as clear in UWP apps as it is in other applications).

I haven't thought about proper solutions for this and I understand it doesn't have high priority. Just wanted to share my thoughts. If I come up with something better I might open a Pull Request. 🙂

PR #239, opened 2019-03-22
set `TERM` to `xterm-256color`

DRSDavidSoft:
### Summary
This PR changes `TERM` from `xterm` to `xterm-256color` to be better consistent with other xterm.js-based Terminals

### Related
- **Hyper:** https://github.com/zeit/hyper/blob/571231e4438cae4eec3e88fa4aab398b5b269433/app/session.js#L32-L36
- **Terminus:** https://github.com/Eugeny/terminus/blob/9e81f0aa0e7274fdddc110ea9c77ca42f500896e/terminus-terminal/src/services/sessions.service.ts#L96-L101


felixse:
Looks reasonable, thanks 👍

PR #245, opened 2019-03-25
Use idiomatic JS import style in web client

hchokshi:
Add-ons imported as per Xterm documentation recommendations: https://xtermjs.org/docs/guides/using-addons/

Importing from dist in JS is implementation dependent - it is better to use the `export`ed paths, as this is the externally exposed API from the module.


felixse:
Thanks 👍

PR #251, opened 2019-03-25
Add convert line endings to CR option

ivanjh:
I've been unable to use multi-line paste in interactive terminal programs (e.g. nano editor) as it expects to see the [Enter] key as [CR].
Sending pasted line-endings as [CR] also still works for all my non-interactive situations, as the default WSL Ubuntu 16/18 `stty icrnl` configuration will convert [CR] to [LF].

### DIFF
--- a/FluentTerminal.App.ViewModels/ShellProfileViewModel.cs
+++ b/FluentTerminal.App.ViewModels/ShellProfileViewModel.cs
@@ -208,6 +208,7 @@ namespace FluentTerminal.App.ViewModels
                     _lineEndingStyle = value;
                     RaisePropertyChanged();
                     RaisePropertyChanged(nameof(DoNotModifyIsSelected));
+                    RaisePropertyChanged(nameof(ToCRIsSelected));
                     RaisePropertyChanged(nameof(ToCRLFIsSelected));
                     RaisePropertyChanged(nameof(ToLFIsSelected));
                     _editingLineEndingStyle = false;
@@ -231,6 +232,12 @@ namespace FluentTerminal.App.ViewModels
         {
             get => LineEndingStyle == LineEndingStyle.ToLF;
             set => LineEndingStyle = LineEndingStyle.ToLF;
+        }
+
+        public bool ToCRIsSelected
+        {
+            get => LineEndingStyle == LineEndingStyle.ToCR;
+            set => LineEndingStyle = LineEndingStyle.ToCR;
         }
 
         public void SaveChanges()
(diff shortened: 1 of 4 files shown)


ericcornelissen:
I'm having trouble compiling and running your changes. It seems `ShellProfileViewModel.cs` is missing a property for `ToCRIsSelected`? Adding the snippet below makes it work for me...

```c#
public bool ToCRIsSelected
{
  get => LineEndingStyle == LineEndingStyle.ToCR;
  set => LineEndingStyle = LineEndingStyle.ToCR;
}
```

Also, a small comment on the design, I would expect the options to be order as ⬇️ but that may just be me

```
1. Do not modify line endings
2. Convert line endings to CR+LF
3. Convert line endings to CR
4. Convert line endings to LF
```

ivanjh:
Sorry. I'd missed a function - updated.
Happy to change order - I was just going with alpha sort.

ivanjh:
@ericcornelissen Do you know much about that property (and the others)?

The pattern of a setter than does not inspect the provided value smells.
```cs
  set => LineEndingStyle = LineEndingStyle.ToCR;
```
I've noticed the line-endings option doesn't always switch when selecting different profiles, and it appears that this is part of it. When selecting other profiles `DoNotModifyIsSelected` property is updated to `False` but the setter doesn't inspect that, and just sets `LineEndingStyle = LineEndingStyle.DoNotModify`.

I'm debugging a little more now.

ericcornelissen:
Not too much to be honest @ivanjh, the original implementation was by @Riebart in #172 and I guess @felixse can tell you more about the design. Neither do I use the feature so can't help on that front either...


> I've noticed the line-endings option doesn't always switch when selecting different profiles, and it appears that this is part of it. When selecting other profiles DoNotModifyIsSelected property is updated to False but the setter doesn't inspect that, and just sets LineEndingStyle = LineEndingStyle.DoNotModify.

Not entirely sure what you are describing, but I would expect changing the `LineEndingStyle` property in the WSL profile doesn't affect the other profiles?

ivanjh:
@ericcornelissen The issue is when you select different profiles (not edit), the UI doesn't always update to show the change in LineEndingStyle (although it appears to save/use the correct values).

Let's not drag that bug into this, for now I've just used consistent code, I'll raise a separate issue/pr.

ericcornelissen:
> @ericcornelissen The issue is when you select different profiles (not edit), the UI doesn't always update to show the change in LineEndingStyle (although it appears to save/use the correct values).

Ah I see, can confirm I'm having the issue as well. Good catch 👍

felixse:
Thanks 👍

Issue #256, opened 2019-03-25
Line ending style in profile preview is not always correct


ericcornelissen:
> I've noticed the line-endings option doesn't always switch when selecting different profiles, and it appears that this is part of it. When selecting other profiles `DoNotModifyIsSelected` property is updated to `False` but the setter doesn't inspect that, and just sets `LineEndingStyle = LineEndingStyle.DoNotModify`.

_Originally posted by @ivanjh in https://github.com/felixse/FluentTerminal/pull/251#issuecomment-476201715_

* * *

Steps to reproduce (approximate):

1. Create a profile with `LineEndingStyle` do not modify
1. Create a profile with `LineEndingStyle` to CR+LF
1. Switch between the profiles a couple of times
1. Both profiles will display the same `LineEndingStyle` in the preview (for me it is always "do not modify")

note that this does not reflect the actual value, if you click edit on the profile with the wrong value, the setting will appear correctly.

ivanjh:
Proposed fix in #257

PR #257, opened 2019-03-25
Fix shell profile radio button line-ending binding

ivanjh:
Addresses #256
Only change when updated from a checked radio button

### DIFF
--- a/FluentTerminal.App.ViewModels/ShellProfileViewModel.cs
+++ b/FluentTerminal.App.ViewModels/ShellProfileViewModel.cs
@@ -199,7 +199,7 @@ namespace FluentTerminal.App.ViewModels
         public LineEndingStyle LineEndingStyle
         {
             get => _lineEndingStyle;
-            set => _lineEndingStyle = value;
+            set => Set(ref _lineEndingStyle, value);
         }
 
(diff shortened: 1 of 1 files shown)


ivanjh:
I have no zero experience in UWP UI binding.
Many hours of digging showed that binding an Enum to a Radio Button isn't as obviously easy as it should be.
I'll need some help verifying this is a reasonable approach.

ivanjh:
Also added fix 75aba93 for existing bug when cancelling profile edit

felixse:
There was a minor problem where the non-edit view was not updated on save, but that was just a missing call to Set() to raise the NotifyPropertyChanged event. Very good solution, we should implement this approach for our other radiobuttons as well 👍

ivanjh:
> Very good solution, we should implement this approach for our other radiobuttons as well 👍
@felixse I did have a quick look (with my limited UWP experience) at other radio buttons in use at the time. They don't seem to be affected in the same way, as we don't update the model beneath them after initialization. It appeared to be this update (when switching the selected profile setting) that was causing the non-selected buttons to post-back their state.

felixse:
just to keep things consistent 😃

PR #258, opened 2019-03-26
added test for DialogService.ShowInputDialogAsync

Meir017:


felixse:
Good catch, thank you 👍

PR #265, opened 2019-03-28
Update Choco install docs again

pgalbraith:
As of v0.4.1.0 user's shouldn't need to manually enable developer mode any more prior to installing with Chocolatey, the install script will turn this on automatically.


skipmeister123:
We should be very careful about this, as there are security implications. 
We should be sure that we are very clear about what we are doing.

ericcornelissen:
That is a good point @skipmeister123. It's interesting that Windows/Chocolatey let's you do it programmatically without the user giving explicit consent.

pgalbraith:
Hey guys, while I fully acknowledge (and largely share) your concerns here, this PR is simple documentation of decisions made elsewhere (i.e. at chocolatey.org), that really don't require direct involvement of @felixse .

In order to handle the concerns you're raising here, I'd recommend moving the discussion to https://chocolatey.org/packages/fluent-terminal.

felixse:
Yes, no point in discussing this here, since it's already done this way on chocolatey.
Guess I should stop procrastinating and get my code signing certificate approved.

PR #277, opened 2019-03-31
[WIP] [Help Wanted] Native Renderer

felixse:
Made basic implementations with 2 different parsers:
First one is based on [VtNetCore](https://www.nuget.org/packages/VtNetCore/)
Second one is using the parser of [teds terminal](https://github.com/IntelOrca/teds-terminal)

Both are using [Win2D](https://github.com/Microsoft/Win2D) for GPU accelerated rendering.

VtNetCore seems to have the more elaborated parser, but no luck so far getting it to work with ConPTY.
The teds terminal parser on the other hand is pretty basic, but performed reasonable well.

Another approach I would love to try is porting the xterm.js parser to c#. This might provide more compatibility to the existing parser/renderer. But this will be hard.

Looking forward to your input. Would be great if someone is willing to join contributing on this, since this will be a lot of effort.

### DIFF
--- a/FluentTerminal.App/FluentTerminal.App.csproj
+++ b/FluentTerminal.App/FluentTerminal.App.csproj
@@ -154,6 +154,26 @@
     <Compile Include="Dialogs\CreateKeyBindingDialog.xaml.cs">
       <DependentUpon>CreateKeyBindingDialog.xaml</DependentUpon>
     </Compile>
+    <Compile Include="Views\NativeFrontend\Ansi\AnsiParser.cs" />
+    <Compile Include="Views\NativeFrontend\Ansi\C0.cs" />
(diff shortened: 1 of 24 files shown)


jerch:
@felixse Is C# mandatory? There are several C++ parser libs out there that look promising, not sure about their .NET support though.

> Another approach I would love to try is porting the xterm.js parser to c#. This might provide more compatibility to the existing parser/renderer. But this will be hard.

Well that could be done, the parser and the terminal buffer are operating on memory/byte level (very C-ish), thus porting might be straight forward. The hard part is what plugs the final terminal abstraction together with tons of JS semantics. Imho this would have to be rewritten from scratch.

felixse:
Yeah, I don't think I can find the time for this. Current plan is to stick with xterm.js and hope that WebView2 (chromium based) will be released before you drop the canvas renderer.

Issue #192, opened 2019-01-31
Add option to notify that text has been copied


felixse:
As @skipmeister123 suggested in #65 we should add an option to show a little notification (similar to the resize info) that text has ben copied to the clipboard

AndrewKralovec:
@felixse , should we reuse the re-size notification or create a separate one ?  The logic would stay mostly the same.

skipmeister123:
If it was me doing the work, I'd want a general "notification" class, which I can invoke with a message (and maybe severity for an icon?).  That way, we open ourselves up to things like visual bells and stuff.

AndrewKralovec:
Would you leave the notifications hooked up to the ViewModel, and then create class to do the logic or would you create an custom control? 

If its the former, does that class go under services dir or should it kept in the view model dir ?

skipmeister123:
As a user, I don't care.  I would say that we may want a class, so that we could have multiple notifications (a bell chimes as you're resizing the window, for example.)

AndrewKralovec:
Okay, I'll give this task a shot as my second ever commit. I moved the Overlay logic into a Overlay class,(since we already use a notification class for the toast). Once i figure out why the view model is not binding to the overlay class attributes, it'll be easy to extend it to show an overlay for the copied text and whatever else its needed for. 

https://github.com/AndrewKralovec/FluentTerminal/tree/DEV/CopiedTextOverlay

felixse:
You are having problems with the bindings because your properties are not emitting the NotifyPropertyChangedEvent. I think the Overlay should actually be a ViewModel, that way you can inherit from ViewModelBase, then you can use the Set(ref _field, value) method to update the underlying field and raise the event.

AndrewKralovec:
@felixse , Yeah i believe you are correct. I was thinking the same after i developed out the class. I appreciate the input. Most of my experience is in ASP.NET Core, and i hope to become better at Universal apps contributing to the project. 

I don't however, think it would be a good idea to repeat the overlay control elements for different overly information. I'll have to figure out how to pass the resize values to the control via an event, and not pass them as fields, otherwise whats the point of creating a control. You could just rename the resize methods to something generic and just changing the text for the desired overlay, be it the resize text, copied text, etc.  

Once again thanks for the responses, @felixse , @skipmeister123. This is great project and I appreciate the ability to contribute.

PR #278, opened 2019-04-01
Dev/overlay control

AndrewKralovec:
This PR is to resolve issue [192](https://github.com/felixse/FluentTerminal/issues/192). I created an Overlay Control & ViewModel to handle any text overlay messages needed on the terminal view. Currently this is used for two events, the terminal resize and the copied text. 

image

image


FluentTerminal.App.ViewModels/TerminalViewModel.cs
```diff
@@ -335,6 +309,7 @@ private async void Terminal_KeyboardCommandReceived(object sender, string e)
                     {
                         var selection = await Terminal.GetSelectedText().ConfigureAwait(true);
                         ClipboardService.SetText(selection);
+                        Overlay.OverlayContent = "Text copied";
```
felixse: I would make this configurable

FluentTerminal.App.ViewModels/OverlayViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public bool ShowOverlay
+        {
+            get => _showOverlay;
+            set
+            {
+                Set(ref _showOverlay, value);
+                if (value)
+                {
+                    if (_overlayTimer.IsEnabled)
```
felixse: It's a bit too much logic for a setter. I think the usage would be easier to understand if there was a method `Show(string message)` which sets our values and starts the timer. We might also make the setters private to avoid confusion

FluentTerminal.App.ViewModels/TerminalViewModel.cs
```diff
@@ -197,6 +171,12 @@ public TerminalTheme TerminalTheme
             set => Set(ref _terminalTheme, value);
         }
 
+        public OverlayViewModel Overlay
```
felixse: An AutoProperty should be enough here since this is only set once in the constructor

FluentTerminal.App.ViewModels/OverlayViewModel.cs
AndrewKralovec: I was thinking since every time i set the value, i want to dispatch the timer, i thought it would be more efficient to keep it in the set. However, i see what you mean and thus fixed this in the newest commit. I now call Show() instead.

FluentTerminal.App.ViewModels/TerminalViewModel.cs
AndrewKralovec: Fix, its now a getter/setter.

FluentTerminal.App.ViewModels/TerminalViewModel.cs
AndrewKralovec: Can you elaborate ?

felixse: Not everybody might want to see this. I would add a toggle to the settings to activate/deactivate this

AndrewKralovec: I see now, yeah that is a good idea, Ill get on that.

AndrewKralovec: @felixse , i was thinking about this feature. Instead of just adding a single toggle to the settings, would it be a good idea to create a list of possible overlay settings/events? The idea would be, create a list in the settings page. The list will have the possible overlay notifications (resize & text copied). This list will be bound to the terminal settings. You can toggle them on and off. In the Terminal View Model, we would pass this list to the overlay view model. Then continue to hook up the overlay events to the view, but instead of just sending a string in the show overlay, it would send an object, such as a dictionary of some sorts specifying the text and the settings name. In the overlay model, it would check the settings name to see if its in the list of events that are toggled on. 

I was thinking this would make it more extensible, but if this is going to be a small feature, then it over kill. 

Example/ proof of the concept.
```

public Dictionary<string, bool> OverlaySettings;
public TerminalPageViewModel()
{
    ... 
    OverlaySettings.Add("resize", false);
    OverlaySettings.Add("copied", false);
}


<StackPanel Margin="{StaticResource ItemMargin}">
    <TextBlock Margin="0,0,0,8" 
                Text="Overlay Notifications" />
    <ListView x:Name="listView1"
                ItemsSource="{x:Bind ViewModel.OverlaySettings}">
        <ListView.ItemTemplate>
            <DataTemplate >
                <ToggleSwitch
                    Margin="{StaticResource ItemMargin}"
                    Header="{Binding Key}"
                    IsOn="{Binding Value, Mode=TwoWay}" />
            </DataTemplate>
        </ListView.ItemTemplate>
    </ListView>
</StackPanel>

private void Terminal_SizeChanged(object sender, TerminalSize e)
{
    // Or whatever type is decided 
    Overlay.Show("resize", $"{e.Columns} x {e.Rows}");
    Overlay.Show(new { Name = "resize", Value = $"{e.Columns} x {e.Rows}" });
    OverlayEvent.?Invoke(this, $"{e.Columns} x {e.Rows}"); 
}
```

felixse: definitively possible, but I don't think there will be much more possible overlays. Maybe keep this in mind until a third overlay comes around?

PR #284, opened 2019-04-02
Native Renderer Collabate PR

AndrewKralovec:
This is my current work as of now. My current task is trying to get the scroll bar functional.


FluentTerminal.App/Views/NativeFrontend/TerminalControl.xaml.cs
```diff
@@ -424,14 +457,19 @@ private void ProcessTerminalCode(TerminalCode code)
...
+  
             var text = Encoding.UTF8.GetString(e, 0, e.Length);
             var reader = new ArrayReader<char>(text.ToCharArray());
             var codes = _parser.Parse(reader);
             foreach (var code in codes)
             {
                 ProcessTerminalCode(code);
             }
-            canvas.Invalidate();
+
```
felixse: we might need to wrap this in a dispatcher. You can see some dispatcher usage in XtermView

Issue #203, opened 2019-02-13
[Suggestion] New logo


skipmeister123:
I've been working on an idea for an updated logo, using the existing one as a base.  Feedback?
![48x48](https://user-images.githubusercontent.com/1911919/52744453-bf2f3a80-2faa-11e9-954d-a7a6064c76b9.png)
![1280x1280](https://user-images.githubusercontent.com/1911919/52744452-be96a400-2faa-11e9-90c5-e60886be4862.png)

ericcornelissen:
Can I ask what your motivation for this change is? It's not clear to me what the additional line is supposed to represent 🤔

skipmeister123:
Sure; fair question.  Here was my thought process in designing it:

The current logo accurately represents the product, however I believe that visibility could be improved.  As it stands, with just a white logo it will get lost on a white background.  Including a framing device will naturally give it contrast.  That way, on a white background, you wouldn't have to change the color - and therefore the branding - of the logo.

The line on the bottom has a few purposes:

1) It offers a suggestion of depth.  It could be viewed as the side of a 3D box, suggesting FluentTerminal has depth.
2) It is representative of the tab bar.
3) It acts as a form of punctuation; and "underline", if you will.  It's a subtle way of saying, "This is your terminal."

Does that help?

--
Hans Kokx
On 2/14/2019 3:19:12 AM, Eric Cornelissen <notifications@github.com> wrote:
Can I ask what your motivation for this change is? It's not clear to me what the additional line is supposed to represent 🤔

felixse:
I'm not a fan of the black background. I would like to keep the icon a single colored shape to be in line with the rest of the Windows 10 icons. 

I think we should make the prompt symbol more unique, for example by adding some perspective to it (similar to the current Visual Studio icons) or something

ericcornelissen:
I agree with @felixse, but I think your argument (contrast and recoloring) is pretty strong @skipmeister123. Even with your explanation I still don't like the line at the bottom though 😅 but that's pure personal preference


> I think we should make the prompt symbol more unique, for example by adding some perspective to it (similar to the current Visual Studio icons) or something

I expect it will be difficult to do that without looking to much like the Visual Studio icons, and I don't think it is a good idea to have logo that people will associate with that brand if it's not actually related.

skipmeister123:
I'm going to think on this and come up with a second proposal.


--
Hans Kokx
[...]

PR #290, opened 2019-04-03
Add dark and light mode SVG icons for #203

mikemaccana:
I use convertio to generate PNGs - you might prefer another tool or different icon sizes though.


felixse:
Visual Studio comes with a good resizing assistent for UWP, but thanks anyway 👍

PR #291, opened 2019-04-03
Un-swap Ubuntu yellow and cyan

MinchinWeb:
The build-in Ubuntu profile has the yellows and cyans swapped. This fixes that.


felixse:
Thanks 👍

PR #293, opened 2019-04-04
Ap/ssh Implemented SSH by using the exsting infrastructure

peske:
I will also call Felix to review this PR, if he has time.


FluentTerminal.App.ViewModels/MainViewModel.cs
```diff
@@ -214,6 +218,34 @@ public Task AddConfigurableTerminal()
...
+                {
+                    if (Terminals.Count == 0)
+                    {
+                        await ApplicationView.TryClose();
+                    }
+
+                    return;
+                }
+
+                var profile = new ShellProfile
```
mjs: Eventually we probably want to define a dedicated `RemoteProfile` type but this will probably do for now

FluentTerminal.App.ViewModels/MainViewModel.cs
```diff
@@ -214,6 +218,34 @@ public Task AddConfigurableTerminal()
...
+                {
+                    Arguments = $"-p {connectionInfo.Port:#####} {connectionInfo.Username}@{connectionInfo.Host}",
+                    Location = @"C:\Windows\System32\OpenSSH\ssh.exe",
```
mjs: Do all versions of Windows which FluentTerminal supports include OpenSSH? If not, there should probably be some graceful handling of this (i.e. explain to the user why it doesn't work).

FluentTerminal.App.ViewModels/MainViewModel.cs
felixse: Yes, that's just a temp solution for now. I think we will need to extend this once we allow to also store connections

FluentTerminal.App.ViewModels/MainViewModel.cs
felixse: Seems like openssh.exe was added in 1709, which is also our current minimum version

peske: OpenSSH is a "Windows Feature", and can be installed if already not. (The same as with WSL.) But yes, I agree that we should handle this! Will create an issue for this one.

Issue #283, opened 2019-04-02
Command line configuration of fluent terminal


darkliquid:
Is there a way I can configure fluent terminal from the command line? I'm trying to script the bootstrapping of my dev environment and would like to be able to import a theme, set it as the active theme and tweak a number of other config settings without having to actually manually open the app and do so.

felixse:
so far not, but our command line parser is quite easy to extend 😉

darkliquid:
In that case, is there a schema for all the settings that are currently supported? 

I was looking at https://github.com/felixse/FluentTerminal/blob/master/FluentTerminal.App.Services/Implementation/SettingsService.cs to try and figure out how the config works (and it looks like it is being stored as JSON in some cases, presumably encoded into the settings.dat file, basd on all the `WriteValueAsJson` calls in there). 

I was thinking about trying to add an `--import-config myconf.json` option that would allow importing all of the config from outside the app, but it's a little hard to know where to start, since I've never written a windows app before, nor worked with C#.

Any pointers welcome :)

darkliquid:
Hmm, llooking at this: https://github.com/felixse/FluentTerminal/blob/master/FluentTerminal.App/App.xaml.cs#L230 am I understanding things correctly that if a `config.json` is placed in the _logging_ directory for fluent, it will import that?

felixse:
That file is actually just configuring serilog, our logging framework.

Something to import a json config file sounds like a good approach. You would need to deserialize it using json.net (you can find something very similar in the theme parser) and pass the values to the settings service.

darkliquid:
Ahh, cool. I shall have a poke around and see what I can come up with :)

PR #296, opened 2019-04-06
feat(cli): Add json import/export of app config

darkliquid:
Add the capability to export and import the application config to/from a json file. The JSON file is LocalState\config.json in the app package folder. The options are available via the new flags to the settings command, e.g. `fluent.exe settings --export` and `fluent.exe settings --import`

Import attempts to merge imported config with existing config, and refuses to import thigns marked as pre-installed. It attempts to merge configuration for pre-installed profiles where those fields are editable, but otherwise discards invalid configuration.

Addresses: https://github.com/felixse/FluentTerminal/issues/283


FluentTerminal.App/App.xaml.cs
```diff
@@ -160,9 +160,72 @@ protected override void OnActivated(IActivatedEventArgs args)
...
+                            var stream = await exportFile.OpenStreamForWriteAsync();
+                            JsonSerializer s = new JsonSerializer();
+                            using (StreamWriter sw = new StreamWriter(stream))
+                            using (JsonWriter writer = new JsonTextWriter(sw))
+                            {
+                                var config = new
+                                {
+                                    App = _settingsService.GetApplicationSettings(),
+                                    KeyBindings = _settingsService.GetCommandKeyBindings(),
+                                    Themes = _settingsService.GetThemes(),
```
felixse: We should avoid exporting preinstalled themes and profiles

FluentTerminal.App/App.xaml.cs
```diff
@@ -160,9 +160,72 @@ protected override void OnActivated(IActivatedEventArgs args)
...
+                                    Profiles = _settingsService.GetShellProfiles(),
+                                    TerminalOptions = _settingsService.GetTerminalOptions(),
+                                    TabThemes = _settingsService.GetTabThemes()
```
felixse: TabThemes are static for now, we don't need to export/import them

FluentTerminal.App/App.xaml.cs
```diff
@@ -160,9 +160,72 @@ protected override void OnActivated(IActivatedEventArgs args)
...
+                                _settingsService.SaveApplicationSettings(config.App);
+                                
+                                foreach (var pair in config.KeyBindings)
+                                {
+                                    _settingsService.SaveKeyBindings(pair.Key, pair.Value);
+                                }
+
+                                foreach (var theme in config.Themes)
+                                {
+                                    _settingsService.SaveTheme(theme, _settingsService.GetTheme(theme.Id).Equals(default(TerminalTheme)));
```
felixse: If we only export non preinstalled themes/profiles we can hardcore IsNew to true

darkliquid:
Pushed an update that hopefully makes a bit more sense - it refuses to operate on `Preinstalled=true` themes and profiles now for both export and import. I'm also trying to make it a little friendlier to merging with existing config, rather than completely replacing it with a fusion of default settings and whatever you happen to have imported.

felixse:
Thanks 👍 
I ahve moved the logic to the settings and fixed some issues. Next I'm gonna write some tests and then this can get merged

darkliquid:
Nice!

PR #299, opened 2019-04-06
Read whole terminal frame at once to avoid console data append glitches

MaxRis:
Fixes console output rendering [issues ](https://github.com/jumptrading/FluentTerminal/issues/13) that appear on whole buffer frame update (happens on usage of tmux, mosh).

More details:

- Similar issues happen both for ConPTY and WinPTY on running `tmux`
- Issue disappears if to increase output std reading buffer. But if to resize the terminal to exceed the buffer the issue come back again. 
- "Unexpected" characters appear between calls of `_appServiceConnection.SendMessageAsync(CreateMessage(e));`
2019-04-07_1-30-41-console-message-issue

Changes in PR read output console stream for whole terminal buffer at once to execute `_appServiceConnection.SendMessageAsync` only single time per "frame".

### DIFF
--- a/FluentTerminal.SystemTray/Services/ConPty/ConPtySession.cs
+++ b/FluentTerminal.SystemTray/Services/ConPty/ConPtySession.cs
@@ -11,6 +11,7 @@ namespace FluentTerminal.SystemTray.Services.ConPty
         private TerminalsManager _terminalsManager;
         private Terminal _terminal;
         private bool _exited;
+        private TerminalSize _terminalSize;
 
         public int Id { get; private set; }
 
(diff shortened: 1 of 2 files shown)


FluentTerminal.SystemTray/Services/ConPty/ConPtySession.cs
```diff
@@ -80,7 +83,8 @@ private void ListenToStdOut()
                 {
                     do
                     {
-                        var buffer = new byte[1024];
+                        int bufferSize = 1024 > _terminalSize.Columns * _terminalSize.Rows * 4 ? 1024 : _terminalSize.Columns * _terminalSize.Rows * 4;
```
felixse: I would suggest Math.Max to make this more readable

MaxRis: @felixse sounds great, thank you! Have updated it

jackluo923:
Is it only me that this merge introduced a bug where upon disconnecting from tmux, the tmux's "bar" at the bottom stays visible until the window is resized?

MaxRis:
@jackluo923 does it happen for you under WSL only? Seems under ssh or mosh the issue is not reproducible

jackluo923:
@MaxRis Thanks for making me aware there are "remote tabs" where we can use ssh or mosh directly. Previously, I have always started WSL in order to ssh and mosh into remote machines. It seems using mosh or ssh directly via FluentTerminal works flawlessly. 

So to answer your question, yes this only happens under WSL.

PR #311, opened 2019-04-09
F11 to toggle fullscreen

TomerZeitune:
Hey, I added F11 default keybinding to toggle fullscreen mode.

### DIFF
--- a/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
+++ b/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
@@ -260,7 +260,7 @@ namespace FluentTerminal.App.Services.Implementation
                         Shift = false,
                         Key = (int)ExtendedVirtualKey.Enter
                     },
-                      new KeyBinding
+                    new KeyBinding
                     {
                         Command = nameof(Command.ToggleFullScreen),
                         Ctrl = false,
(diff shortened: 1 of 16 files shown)


skipmeister123:
Why were we lacking this in the first place? 😅 Thank you!

felixse:
@mjs whats your opinion on this, regarding #309?

mjs:
It's probably OK. gnome-terminal uses F11 for full screen too, so this key is probably not widely used by console apps.

FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
```diff
@@ -259,6 +259,14 @@ public ICollection<KeyBinding> GetDefaultKeyBindings(Command command)
                         Alt = true,
                         Shift = false,
                         Key = (int)ExtendedVirtualKey.Enter
+                    },
+                      new KeyBinding
+                    {
+                        Command = nameof(Command.ToggleFullScreen),
+                        Ctrl = false,
+                        Alt = false,
+                        Shift = false,
```
mjs: There's no need to put in the `false` values. This is the default if not specified and it's easier to read the definition without them.

felixse:
Thanks 👍

PR #325, opened 2019-04-18
Ap/key bug

peske:
I believe that this should fix the bug with Appveyor.


peske:
Fuck, wrong repo! Deleting. Sorry!

PR #330, opened 2019-04-19
added tests for KeyboardCommandService.DeregisterCommandHandler

Meir017:


Meir017:
I doesn't really matter what the command text is, I just used the same logic that's used in the other tests for generating the command

ericcornelissen:
My apologies, never looked at the other tests 😅 In that case my commend applies to the whole test suite rather than your addition 👍

Issue #310, opened 2019-04-09
Allow multiline pasting with `\` in WSL


manuhornung:
Maybe this is in fact two issues, but as I'm not sure yet I just opened this issue for now.

## The Problem
I want to paste a multi-line command into my WSL terminal running Ubuntu 18.10
e.g.:
```
sudo apt-get install -y \
    apt-transport-https \
    ca-certificates \
    curl \
    software-properties-common
```
### Expected behavior
The multi-line text is pasted as a single command and I can then press return to execute it.

### Actual behavior
The multi-line text is pasted as one command per line. Every line but the last one are executed as separate commands. The last one ends up in the input line where I can then press return to execute it (but which will break since it's not a complete command).

### Additional notes
- I can manually enter the same multi-line command without issues
- This might have something to do with how the pasting is handled under Windows. I know that PowerShell uses `` ` `` for marking line ends in a multi-line command while cmd uses `^` and Linux uses `\`. Maybe we need to do some intermediate replacement with these and then back for pasting?
- I fiddled around with all possible settings for line end replacements (`CR/LF` etc.). That didn't change anything except for adding additional blank lines in between the commands for `CR+LF` and `Do not modify line endings`.
- Pasting multi-line commands delimited with `` ` `` or `^` works in PowerShell or cmd tabs, respectively
- Pasting the command directly into the Ubuntu WSL shell without FluentTerminal also works as expected

alexjmoore:
@manuhornung as per my PR, can you test changing your "Line ending translation mode" under Settings -> Profiles -> WSL to "Convert to LF", save and then re-launch a WSL terminal and test a multi-line copy-paste again?

felixse:
This should work now thanks to #557

PR #348, opened 2019-05-01
Updated default line ending for WSL to address issue #310

alexjmoore:
As per the subject, I believe changing the line style ending default for WSL should address issue #310.  Unix terminals typically expect a LF ending and windows will default generally to CR+LF.  At least in my testing changing to LF fixes the multi-line copy paste issue, so suggest changing the default for WSL.


felixse:
Good idea, thanks 👍

PR #350, opened 2019-05-03
Keep working directory

zuozishi:
After opening the Fluent Terminal with the right-click menu, the new terminal remains in the working directory.


FluentTerminal.App.ViewModels/MainViewModel.cs
```diff
@@ -269,6 +269,14 @@ public void AddTerminal(Guid shellProfileId)
 
         public void AddTerminal(ShellProfile profile)
         {
+            if(Terminals!=null && Terminals.Count > 0)
+            {
+                var currentProfile = Terminals[0].ShellProfile;
+                var originProfile = _settingsService.GetShellProfile(currentProfile.Id);
+                if(originProfile.WorkingDirectory != currentProfile.WorkingDirectory)
```
AndrewKralovec: I would make this configurable in the settings.

felixse:
I'm not sure if this is a good idea. This will only capture the initial directory of the current tab, but any navigation is not considered. I'm afraid this might be more confusing than it helps.

skipmeister123:
I agree that this is a bad idea. What happens if you're in a WSL path and 
open CMD? Or you're in a UNC path and open a WSL?


I think a better solution to your problem would be to have a cascading 
right click menu in explorer which gives you a choice of profiles to open 
the path in.

felixse:
This is partly solved by #489 which allows to duplicate tabs. I'm closing this for now due to conflicts and the issue with different profiles @skipmeister123 described

PR #357, opened 2019-05-07
Leave tabs open on error

mjs:
Do not close tab if shell process closes with an error. An indicator is shown in the tab's title when this happens.

Example:

image


igorgatis:
Tab is kept when `exit` has no parameter:

image

It does quit with `exit 0` though.

PR #387, opened 2019-05-30
Disable context menu for TextBox controls of SSH connection dialog

MaxRis:
Fixes https://github.com/jumptrading/FluentTerminal/issues/86

Workaround to disable non-customizable default context menu for TextBox control.

### DIFF
--- a/FluentTerminal.App/Dialogs/SshInfoDialog.xaml
+++ b/FluentTerminal.App/Dialogs/SshInfoDialog.xaml
@@ -7,6 +7,7 @@
     xmlns:vm="using:FluentTerminal.App.ViewModels"
     xmlns:controls="using:Microsoft.Toolkit.Uwp.UI.Controls"
     xmlns:enums="using:FluentTerminal.Models.Enums"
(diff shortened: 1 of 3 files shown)


felixse:
image
Cannot reproduce it (Windows 10 1903)
Also, do we really need to disable it just because its color scheme is wrong? Seems a bit drastic 🤔

MaxRis:
@felixse thank you for checking, I've tried on 1809. Will double check on 1903 as well.

> Also, do we really need to disable it just because its color scheme is wrong? Seems a bit drastic

I was not able to find a way to customize the style/theme for default context menu. Alternative is to implement custom context menu, but the question was if it really needed there.

felixse:
Why can't we just leave it as it is? I mean, it's not causing any issue besides not looking perfect

MaxRis:
"not perfect" it was the mental reason :) @felixse thank you for figuring this out, I can't see the issue on 1903 as well. Seems, MS silently fixed the issue. Going to close the PR.

PR #390, opened 2019-06-02
Implementing remote SSH profiles.

peske:
Many commits from https://github.com/jumptrading/FluentTerminal/tree/ap/remoteprofiles combined in a single commit here. It was easier to do this way than to cherry-pick and merge particular commits.

It can be tested in https://github.com/jumptrading/FluentTerminal/tree/ap/4up-remoteprofiles


mjs:
Remaining issues:
- Address conflicts with master
- When the "New Configurable SSH Window" (Ctrl-Alt-Shift-Y) is used, the SSH Connection Info dialog appears below the new terminal window. This is confusing. Can the creation of the new terminal window  be delayed until after the SSH connection dialog has been accepted?

FluentTerminal.App/App.xaml.cs
```diff
@@ -71,9 +71,9 @@ public App()
                 RoamingSettings = new ApplicationDataContainerAdapter(ApplicationData.Current.RoamingSettings),
                 KeyBindings = new ApplicationDataContainerAdapter(ApplicationData.Current.RoamingSettings.CreateContainer(Constants.KeyBindingsContainerName, ApplicationDataCreateDisposition.Always)),
                 ShellProfiles = new ApplicationDataContainerAdapter(ApplicationData.Current.LocalSettings.CreateContainer(Constants.ShellProfilesContainerName, ApplicationDataCreateDisposition.Always)),
-                Themes = new ApplicationDataContainerAdapter(ApplicationData.Current.RoamingSettings.CreateContainer(Constants.ThemesContainerName, ApplicationDataCreateDisposition.Always))
+                Themes = new ApplicationDataContainerAdapter(ApplicationData.Current.RoamingSettings.CreateContainer(Constants.ThemesContainerName, ApplicationDataCreateDisposition.Always)),
+                SshShellProfiles = new ApplicationDataContainerAdapter(ApplicationData.Current.LocalSettings.CreateContainer(Constants.SshShellProfilesContainerName, ApplicationDataCreateDisposition.Always))
```
felixse: I think those can also be saved to roaming since they don't depend on something installed locally like normal profiles do

FluentTerminal.App/App.xaml.cs
```diff
@@ -345,6 +341,7 @@ protected override async void OnLaunched(LaunchActivatedEventArgs args)
             {
                 await InitializeLogger();
 
+                // TODO: Check the reason for such strange using of tasks!
```
felixse: Didn't implement it, but apparently this is meant to run on another task

FluentTerminal.App/Strings/de/Resources.resw
```diff
@@ -832,4 +832,8 @@ Fuzzy</comment>
   <data name="ConfirmDeleteProfile" xml:space="preserve">
     <value>Sind Sie sicher dass Sie dieses Profil löschen wollen?</value>
   </data>
+  <data name="SshProfiles.Text" xml:space="preserve">
```
felixse: please do not add them manually (they will be overwritten), register as a contributor instead: https://poeditor.com/join/project/TfKNIt826z

FluentTerminal.Models/Extensions.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using FluentTerminal.Models.Enums;
+using Windows.ApplicationModel.Resources;
+
+namespace FluentTerminal.Models
+{
+    public static class Extensions
+    {
+        public static string GetResource(this string resource) =>
```
felixse: this is not really declarative in what it does

FluentTerminal.Models/Extensions.cs
```diff
@@ -0,0 +1,72 @@
...
+            ResourceLoader.GetForCurrentView().GetString(resource.Replace('.', '/'));
+
+        public static SshConnectionInfoValidationResult GetValidationResult(this ISshConnectionInfo sshConnectionInfo,
```
felixse: not a fan of mixing extensions for different types in the same file

FluentTerminal.Models/ShellProfile.cs
```diff
@@ -64,23 +64,29 @@ public string TranslateLineEndings(string content)
         public Guid TerminalThemeId { get; set; }
         public ICollection<KeyBinding> KeyBindings { get; set; } = new List<KeyBinding>();
 
-        public override bool Equals(object obj)
+        public virtual bool EqualTo(ShellProfile other)
```
felixse: what was the problem with .Equals()?

peske:
I'm closing this one in order to create a new one.

felixse:
Would have preferred to review this incrementally. That way I have to review this all over again

FluentTerminal.App/Strings/de/Resources.resw
peske: Ok, sorry.

FluentTerminal.Models/Extensions.cs
peske: It gets a string from the resources, based on the key.

FluentTerminal.Models/Extensions.cs
peske: OK, will split.

FluentTerminal.Models/ShellProfile.cs
peske: it is not recommended to do it partially. If you want to override `Equals()`, you should always override `GetHashCode()`. Only this way `Equals()` will fully remain the expected behavior.

FluentTerminal.Models/Extensions.cs
felixse: `I18N.Translate("foo")` was kind of clear, `"foo".GetResource()` imho not so much

PR #395, opened 2019-06-05
Implementing remote SSH profiles, 2nd attempt

peske:
Comparing to the previous remote SSH profiles PR, changes are: 
- All conflicts resolved;
- new window requested behavior improved (the issue reported by @mjs ).


peske:
Closing to make few more changes, according to @felixse comments.

peske:
Reopening...

FluentTerminal.SystemTray/Services/TerminalsManager.cs
```diff
@@ -1,117 +1,110 @@
...
-        }
-
-        public string GetDefaultEnvironmentVariableString(Dictionary<string, string> additionalVariables)
-        {
-            var environmentVariables = Environment.GetEnvironmentVariables();
-            environmentVariables["TERM"] = "xterm-256color";
-            environmentVariables["TERM_PROGRAM"] = "FluentTerminal";
-            environmentVariables["TERM_PROGRAM_VERSION"] = $"{Package.Current.Id.Version.Major}.{Package.Current.Id.Version.Minor}.{Package.Current.Id.Version.Build}.{Package.Current.Id.Version.Revision}";
-
+﻿using FluentTerminal.Models;
```
felixse: did you change your newlines to lf? I think we should stick to crlf to avoid this in the future

FluentTerminal.App.ViewModels/MainViewModel.cs
```diff
@@ -283,25 +281,25 @@ public async Task AddSavedShhRemoteTerminal()
             }
             else
             {
-                await ApplicationView.RunOnDispatcherThread(() => AddTerminal(profile));
+                await AddTerminalAsync(profile);
```
felixse: are you sure about removing the Dispatcher here? There might be a case were this was necessary

FluentTerminal.App.ViewModels/MainViewModel.cs
```diff
@@ -258,7 +258,7 @@ public async Task AddRemoteTerminalAsync()
             }
             else
             {
-                await ApplicationView.RunOnDispatcherThread(() => AddTerminalAsync(profile));
```
felixse: Also this one

FluentTerminal.Models/Resources.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Windows.ApplicationModel.Resources;
+
+namespace FluentTerminal.Models
+{
+    public static class Resources
+    {
+        public static string GetString(string resourceId) =>
```
felixse: is this only used in `SshConnectionInfoValidationResultExtensions` ? Why not use the already existing `I18N.Translate()`?

FluentTerminal.SystemTray/Services/TerminalsManager.cs
peske: In my case git is configured to "check out as-is" and "commit as-is", meaning that I haven't changed line endings. 
But I've also noticed that someone in our fork (probably @MaxRis ?) has configuration "check-out CRLF" and "commit LF", which is the default git configuration of git in Windows, and which caused some headaches to me also.
Since I've started work from our fork, chances are that I've retained some LF files.
You should raise this in Slack so that everyone adopts.

FluentTerminal.App.ViewModels/MainViewModel.cs
peske: Yes, I'm sure. As you can see if you follow the calls, in the last method in the stack `AddTerminalAsync()` uses the dispatcher.
So far we had a situation when we are calling dispatcher several times for the same task, meaning that we are scheduling dispatcher task that schedules dispatcher task to do something...

peske: At the end of the day, you can test in my branch.

FluentTerminal.App.ViewModels/MainViewModel.cs
peske: The same answer...

FluentTerminal.SystemTray/Services/TerminalsManager.cs
felixse: yeah, maybe a .editorconfig might help

FluentTerminal.Models/Resources.cs
peske: Because I've needed translations in `FluentTerminal.App.Models` project (i have a method which returns error message there), and `I18N` class isn't accessible there (dependence tree). I was thinking about moving it, but since it is only about a single line of code I've simply written it again, to minimize change.

FluentTerminal.SystemTray/Services/TerminalsManager.cs
peske: Again, please share this info in Slack because it isn't my git.

FluentTerminal.App.ViewModels/MainViewModel.cs
```diff
@@ -92,8 +93,8 @@ public class MainViewModel : ViewModelBase
             _applicationSettings = _settingsService.GetApplicationSettings();
             TabsPosition = _applicationSettings.TabsPosition;
 
-            AddLocalShellCommand = new RelayCommand(() => AddTerminal());
-            AddRemoteShellCommand = new RelayCommand(() => AddRemoteTerminal());
+            AddLocalShellCommand = new RelayCommand(async () => await AddTerminalAsync());
```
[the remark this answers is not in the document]
peske: OK, do you want me to change that?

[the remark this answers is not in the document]
felixse: I can do this afterwards, there might be a few more places I should check as well

FluentTerminal.Models/Resources.cs
felixse: Ok, I see. I think I will move this later on

felixse:
Do you have a list of resources I need to add to the translations?

peske:
No. But I'll take a look at changes again and prepare the list.

felixse:
https://github.com/felixse/FluentTerminal/pull/390#discussion_r290408047
this is still open

peske:
Regarding https://github.com/felixse/FluentTerminal/pull/390#discussion_r290408047 - I agree. Somehow I've forgotten about it. Can you change it, or you want me to change?

FluentTerminal.SystemTray/Services/TerminalsManager.cs
MaxRis: @peske I have in global git config `autocrlf = false`, it shouldn't be `git`.
I'm working in Visual Studio 2017 Community and noticed that it always seems replaces edited lines with "CRLF". One or more times, seems, I've committed changes with "CRLF" line endings, but later tried to revert back to LF in separate commits.
To avoid above issue with VS I'm keeping `.editorconfig` file locally and VS always sets LF line endings.
But strange thing seems that some .cs files in the project have all lines with CRLF endings, but others files have LF endings. Should we standardize all files to LF line endings? cc: @felixse

peske: @felixse mentioned in the first comment here that he wants CRLF. Because of such issues, I don't rely on VS git extension but do commits manually. In my case it is "as-is". If the file was CRLF - I'm committing CRLF. If the file was LF - I'm committing LF. This way I avoid causing problems to others.
And I believe that this is the only reasonable config - not let git to mess up...

felixse: rather CRLF, thats the default on Windows

felixse:
Just checked this out and gave it a try, looks very nice 👍 
Only thing I didn't get is 'set default' for ssh profiles. Does this do anything or is it a leftover from the normal profiles?

FluentTerminal.SystemTray/Services/TerminalsManager.cs
peske: All the files created by VS are CRLF anyway, so disabling any _smartness_ of git / VS will be enough.

peske:
Thanks. Regarding 'set default' - let me take a look and I'll answer. It was many days ago...

peske:
'set default' - does nothing. Leftover. For now we should simply remove it. But maybe later we may implement this feautre?

felixse:
Ok, then we should remove it for now. I will merge this now anyways. I won't be here for the next few days and that way it's easier for you guys to continue

peske:
Thanks!

PR #398, opened 2019-06-11
Fix crash when any keybinding is changed

mjs:
This fixes a new regression introduced in 0065827bf (SSH profile support) which causes any keybinding change to crash FT. 

It's not clear to me why these JSON serializer settings were introduced so there might be a better way to fix this.


peske:
I've introduced JSON serializer settings because I had issue serializing `SshProfile` class which inherits from `ShellProfile` class. Will recheck if it still happens, and try to find an alternative solution.

PR #403, opened 2019-06-12
Removing duplcated code from profile view models.

peske:
Before this PR we had situation in which `SshProfileViewModel` and `ShellProfileViewModel` were independent classes, although about 75% of code was overlapping. Now I've introduced inheritance so that the former inherits from the later one, and this way I've significantly removed A LOT of duplicated code from `SshProfileViewModel` code (from 520+ lines to about 140 lines).

In order to do so I've had to slightly change `ShellProfileViewModel` also, to make it more _base-class-friendly_. Although according the differences it looks that this class is significantly changed, it actually isn't the case - the only changes are in reorganizing some code, and extracting few methods from previously inline code. It does the same work as before.

This PR also includes few more minor improvements:
- "Username" box in "SSH Profiles" gets prefilled with the actual users username;
- "Line ending translation" in "Profiles" is changed from radio button list to a drop down list (it simplifies both UI and the view model), and now it's consistent with "SSH Profiles".

Also I need to note that few files are committed with LF line endings because the originals in masters were using LF, so I've avoided false differences this way. Btw. It turned out that 260+ files in felixse/FluentTerminal master branch are still using LF line endings. This should be converted to CRLF (as we agreed with @felixse ) because it causes headaches... Our fork is 100% CRLF now.


peske:
After updating the original branch, this PR also solves another bug: https://github.com/jumptrading/FluentTerminal/issues/119

PR #406, opened 2019-06-15
Normalize line endigns to Windows CRLF format

MaxRis:


MaxRis:
Have missed lines ending normalization for .xaml files. Going to add in separate commit to this PR.

ericcornelissen:
Perhaps it is a extra/alternative/better idea to add [a `.gitattributes` file](https://git-scm.com/docs/gitattributes)?

felixse:
Yes, also thought about this.
I'm not quite sure anymore if changing everything to crlf is the right thing. Most people seem to work with autocrlf=true, and I think the problems occurred because some people had this turned off. Maybe we are better off if we set autocrlf=true in a .gitattributes file

MaxRis:
Sounds good!
Have created another PR to introduce .gitattributes file with appropriate settings

MaxRis:
https://github.com/felixse/FluentTerminal/pull/408

PR #409, opened 2019-06-17
Adding a scrollbar to SSH Info dialog.

peske:
Trivial change which solves https://github.com/jumptrading/FluentTerminal/issues/67


peske:
One more minor change added to the same PR: improving remote SSH profile selection.

PR #410, opened 2019-06-17
Standardise on naming of SSH related functionality

mjs:
In some places code and UI referring to SSH connections and profiles used the word "remote". Standardise on "SSH" (this is used more often anyway).


mjs:
@felixse poke :)

PR #423, opened 2019-06-30
Implemented "Quick SSH" dialog

peske:
"Quick SSH" dialog can be invoked by Ctrl+Shift+Q or Ctrl+Alt+Q.

In order to implement it, I've had to refactor and reorganize view-models infrastructure, and for this reason the number of changes is huge. The key change when it comes to profile view-models can be seen in `FluentTerminal.App.ViewModels`, in `Profiles` directory, and in `ShellProfileViewModel` and `SshProfileViewModel` files (in my branch).

Besides that some additional refactoring is done:
- `FluentTerminal.App\Views\TemplateSelectors.xaml` is created and contains code that was repeated in three different files previously.
- SSH-related implementation is changed so that few files are now removed: `ISshConnectionInfo`, `ISshHelperService`, `SshHelperService`, etc. The code is organized in a more natural way now.

@felixse - Please sorry for such huge PR, but it had to be... Also please add the following translations to the resources:
```
Key										Value
-----------------------------------------------
Command.NewQuickSshTab 					"New quick SSH tab"
Command.NewQuickSshWindow 				"New quick SSH window"
x:Uid="QuickSshDialog" 					"Quick SSH"
InvalidCommand 							"Invalid command."
LocationCantBeEmpty 					"Location cannot be empty."
```


mjs:
This is quite nice. There's obviously more to be done it's a great start. I have looked at this change from the outside so far - I haven't reviewed the code. Some things that I like need to be fixed before this can land:

- single character host names don't seem to be allowed (e.g. `ssh x`) but two characters is (e.g. `ssh x9`. This seems a little arbitrary - please allow single character hostnames
- it's a little unclear what the input field actually accepts. Could the text box placeholder text give an example. Instead of `Enter SSH command to execute` how about something like `Enter SSH command (e.g. "ssh username@host")`

I will write up my thoughts about future enhancements for this feature at https://github.com/jumptrading/FluentTerminal/issues/135

peske:
@mjs - just fixed single character hosts issue. Regarding the other one - we should probably first clarify questions at https://github.com/jumptrading/FluentTerminal/issues/135.

peske:
@felixse , I've slightly simplified this one (combined "SSH Info" and "Quick SSH" into a single dialog).

List of resources needed (you can ignore the previous one):
```
Key										Value
-----------------------------------------------
QuickSshDefault 						"Use quick SSH by default"
InvalidCommand 							"Invalid command."
LocationCantBeEmpty 					"Location cannot be empty."
SshInfoDialog.Command 					"Command"
SshInfoDialog.MoshPortRange 			"Mosh port range"
```

peske:
Here's a short [video demo](https://youtu.be/U2RU-eP6v8w).

FluentTerminal.App.ViewModels/Profiles/ProfileProviderViewModelBase.cs
```diff
@@ -0,0 +1,403 @@
...
+            set
+            {
+                TabTheme theme = value;
+
+                if (theme == null)
+                {
+                    // How to handle attempt to setting null theme?
+                    // 1) Throw an exception:
+                    //throw new ArgumentNullException();
+                    // 2) Ignore the attempt:
```
felixse: I think 2) would be better, since an attempt to set it null will most probably be some binding doing stupid things

FluentTerminal.Models/ShellProfile.cs
```diff
@@ -42,8 +42,8 @@ protected ShellProfile(ShellProfile other)
         public int TabThemeId { get; set; }
         public LineEndingStyle LineEndingTranslation { get; set; }
         public Dictionary<string, string> EnvironmentVariables { get; } = new Dictionary<string, string>();
-
-        public bool UseConPty { get; set; }
+        // Setting ConPTY as default
+        public bool UseConPty { get; set; } = true;
```
felixse: Can we move this to to DefaultValueProvider? That's where the other defaults are stored

felixse: Or is this for adhoc profiles? In that case I would set this where it is actually used

peske: This line is removed completely - it turned out it isn't needed (it's already set in the constructor).

FluentTerminal.App.ViewModels/Profiles/ProfileProviderViewModelBase.cs
peske: Changed to ignore.

peske:
Saving executed commands history additionally implemented, to be used for _smart autocomplete_.

@felixse , can you merge this one? It's needed for adding more features... Or you have some problems with it? If so - pls let me know and I'll fix.

PR #428, opened 2019-07-03
Settings option to mute WinPTY's conhost.

### LINKED ISSUE #386: No beeps
Please disable, or at least add an option, to disable beeps from terminals.
I never want them.

MaxRis:
Fixes #386 

- Global settings option is added to mute terminal beeps (mute state is applied for further launched terminal sessions)
- On time of tray process startup global mute settings is applied
- Default audio device is muted while beep to trigger conhost audio sessions is generated

### DIFF
--- a/FluentTerminal.App.Services/ITrayProcessCommunicationService.cs
+++ b/FluentTerminal.App.Services/ITrayProcessCommunicationService.cs
@@ -35,5 +35,7 @@ namespace FluentTerminal.App.Services
         Task<string[]> GetFilesFromSshConfigDirAsync();
 
         Task<bool> CheckFileExistsAsync(string path);
+
+        void MuteTerminal(bool mute);
     }
 }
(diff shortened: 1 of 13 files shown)


felixse:
Is this also working for ConPty?

MaxRis:
@felixse ConPTY doesn't produce any bell sound at all.

The only exception that I've seen was Windows build of `VIM` that produced sounds via "System Sounds" audio profile. I guess, it happens independently from what pty is used - WinPty or ConPTY.

@felixse should I rebase the changelist to solve merge conflicts?

felixse:
Ok. Rebase is fine by me

MaxRis:
> Is this also working for ConPty?

More info: ConPTY doesn't spawn ConHost process in background as WinPTY does, so ConPty is not covered at all.

MaxRis:
Unfortunately, have found the issue to apply switched mute setting. Will create new PR with updated changelist.

PR #437, opened 2019-07-08
Settings option to mute conhost.

### LINKED ISSUE #386: No beeps

MaxRis:
Fixes #386 

- Global settings option is added to mute terminal beeps (mute state is applied to all currently running conhost processes with accessible audio sessions).
- On time of tray process startup global mute setting is applied.
- Default audio device is muted while manually generated beep to trigger conhost audio sessions is playing.

### DIFF
--- a/FluentTerminal.App.Services/ITrayProcessCommunicationService.cs
+++ b/FluentTerminal.App.Services/ITrayProcessCommunicationService.cs
@@ -35,5 +35,7 @@ namespace FluentTerminal.App.Services
         Task<string[]> GetFilesFromSshConfigDirAsync();
 
         Task<bool> CheckFileExistsAsync(string path);
+
+        void MuteTerminal(bool mute);
     }
 }
--- a/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
+++ b/FluentTerminal.App.Services/Implementation/DefaultValueProvider.cs
@@ -29,7 +29,8 @@ namespace FluentTerminal.App.Services.Implementation
                 UseMoshByDefault = true,
                 AutoFallbackToWindowsUsernameInLinks = true,
                 UseQuickSshConnectByDefault = false,
-                RTrimCopiedLines = true
+                RTrimCopiedLines = true,
+                MuteTerminalBeeps = true
             };
         }
 
--- a/FluentTerminal.App.Services/Implementation/TrayProcessCommunicationService.cs
+++ b/FluentTerminal.App.Services/Implementation/TrayProcessCommunicationService.cs
@@ -138,6 +138,11 @@ namespace FluentTerminal.App.Services.Implementation
             return JsonConvert.DeserializeObject<CommonResponse>((string) responseMessage[MessageKeys.Content]).Success;
         }
 
+        public void MuteTerminal(bool mute)
+        {
+            _appServiceConnection.SendMessageAsync(CreateMessage(new MuteTerminalRequest { Mute = mute }));
+        }
+
         public async Task<CreateTerminalResponse> CreateTerminal(byte id, TerminalSize size, ShellProfile shellProfile, SessionType sessionType)
         {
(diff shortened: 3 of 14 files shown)


MaxRis:
@mjs thank you for checking and testing!

felixse:
Toggling this is causing iTunes/Netflix app to go mute for a 1-2 seconds 🤷‍♂

MaxRis:
@felixse normally user shouldn't toggle the switch.
And yes, whole audio device is muted to not play programmatically generated bell sound.
That programmatically generated bell sound is triggered on every conhost mute action to activate audio session for conhost process.

MaxRis:
We have come up to this workaround after few iterations and discussion there https://github.com/felixse/FluentTerminal/issues/386#issuecomment-505257439

MaxRis:
Basically, it does:

1. Mutes whole device https://github.com/felixse/FluentTerminal/pull/437/files#diff-cd75e42d7c72afd7283483cd9a6cd982R577

2. Generates bell sound to have active conhost audio session (we want avoid user to hear that bell sounds by muting whole device at 1st stage) https://github.com/felixse/FluentTerminal/pull/437/files#diff-cd75e42d7c72afd7283483cd9a6cd982R583

3. Applies required mute state to whole found conhost audio sessions https://github.com/felixse/FluentTerminal/pull/437/files#diff-cd75e42d7c72afd7283483cd9a6cd982R600

4. Unmutes device back https://github.com/felixse/FluentTerminal/pull/437/files#diff-cd75e42d7c72afd7283483cd9a6cd982R612

MaxRis:
P.S. To observe visually the process of muting/unmuting you might want to click "Open Volume mixer" menu appeared on mouse right click on Sounds icon in Windows Tray. Opened "Volume Mixer" window will list all active audio sessions. To observe new audio session you need to open WinPTY based session in Fluent and generate bell sound (For example, send CTRL+G in opened WinPTY session).

cjw296:
Hmm, will "mute whole device" mean that if I'm listening to audio (on a zoom call, music, etc), I'll get a mute dip when this kicks in? If so, I'd probably prefer just to hear the first bell...

MaxRis:
@cjw296 yes, but audio device muting will happen only on start of FluentTerminal.Tray process ( happens only once on first run of Fluent ).

@cjw296 @felixse @mjs 

Possible alternative solution without device muting and with single bell sound might look like:

- Tray process on startup searches for available conhost audio sessions. If no audio session is available, detached cmd.exe process is spawned to generate unmuted bell sound. Tray process searches for conhost audio session again and sets relevant mute state.
- Since detached cmd.exe process spawned on first step with it's active audio session will seat active in background (need to investigate if audio session is not get expired automatically or just disappears for some reasons), Fluent will not generate more bell sounds on next launches or on change of Mute Terminal option value in Settings.

cjw296:
@MaxRis - oh, if it's only when the tray app is first started, that's much less of a problem :-)
I don't know which I'd prefer, an audible sound (maybe do a funky "starting up" noise and make a feature of this? ;-)) or a volume dip across the whole device, but given that it's a one-time thing, whatever works :-)

MaxRis:
@cjw296 okay, thanks! I will create separate Issue to try to enhance current behavior as described above

MaxRis:
Updated https://github.com/jumptrading/FluentTerminal/issues/102

mjs:
> @MaxRis - oh, if it's only when the tray app is first started, that's much less of a problem :-)
> I don't know which I'd prefer, an audible sound (maybe do a funky "starting up" noise and make a feature of this? ;-)) or a volume dip across the whole device, but given that it's a one-time thing, whatever works :-)

Although it would be cool, FT can't make an arbitrary "starting up" sound because the sound needs to come from a conhost.exe process. AKAIK, the only way to do this is to generate a terminal beep.

PR #438, opened 2019-07-08
Settings option to dump terminal output into file.

MaxRis:
Introduces Settings options to
- enable/disable terminal output into log files (disabled by default)
- option to filter out escape sequences to be added into log files (enabled by default)
- path to directory where to create log files (default value is `ApplicationData.Current.LocalCacheFolder.Path` which usually equals to `C:\Users\%USERNAME%\AppData\Local\Packages\53621FSApps.FluentTerminal_87x1pks76srcp\LocalCache`)

Log files are created per terminal session and named to contain session creation timestamp and profile/executable name, for example, `20190708104435232_Powershell.log`.

Feature requested by https://github.com/jumptrading/FluentTerminal/issues/75

### DIFF
--- a/FluentTerminal.SystemTray/Services/TerminalsManager.cs
+++ b/FluentTerminal.SystemTray/Services/TerminalsManager.cs
@@ -1,4 +1,5 @@
-﻿using FluentTerminal.Models;
+﻿using FluentTerminal.App.Services;
+using FluentTerminal.Models;
 using FluentTerminal.Models.Enums;
 using FluentTerminal.Models.Requests;
 using FluentTerminal.Models.Responses;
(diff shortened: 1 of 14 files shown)


mjs:
Thanks. I've tried this out and it works quite well. Filtering of non-printable characters works but there's still plenty of "unexpected" content in there because of the way terminal updates are done. I don't know if there's much that can be done about that.

Potential future improvement: keyboard binding to enable disable recording on a per-terminal basis

FluentTerminal.SystemTray/Services/TerminalsManager.cs
```diff
@@ -8,21 +9,55 @@
...
+                {
+                    string strOutput = System.Text.Encoding.UTF8.GetString(logOutput);
+                    strOutput = EscapeSequencePattern.Replace(strOutput, "");
+                    logOutput = Encoding.UTF8.GetBytes(strOutput);
+                }
+
+                var logFileStream = System.IO.File.Open(_settingsService.GetApplicationSettings().LogDirectoryPath + @"\" + 
+                    _terminals[terminalId].StartTime.ToString("yyyyMMddhhmmssfff") + "_" + _terminals[terminalId].ProfileName + ".log", System.IO.FileMode.Append);
+                logFileStream.Write(logOutput, 0, logOutput.Length);
+                logFileStream.Close();
```
mjs: put this in a `finally` block to ensure the file is always closed?

FluentTerminal.SystemTray/Services/TerminalsManager.cs
```diff
@@ -8,21 +9,55 @@
...
+                var logOutput = output;
+                if (_settingsService.GetApplicationSettings().PrintableOutputOnly)
+                {
+                    string strOutput = System.Text.Encoding.UTF8.GetString(logOutput);
+                    strOutput = EscapeSequencePattern.Replace(strOutput, "");
+                    logOutput = Encoding.UTF8.GetBytes(strOutput);
+                }
+
+                var logFileStream = System.IO.File.Open(_settingsService.GetApplicationSettings().LogDirectoryPath + @"\" + 
+                    _terminals[terminalId].StartTime.ToString("yyyyMMddhhmmssfff") + "_" + _terminals[terminalId].ProfileName + ".log", System.IO.FileMode.Append);
```
mjs: Extract the creation of the path to the log file to its own method to make this more readable

mjs: Given that the StartTime for each session is fixed, can the time string be cached?

FluentTerminal.SystemTray/Services/TerminalsManager.cs
MaxRis: @mjs right, have put it into `using` + `try/catch`

FluentTerminal.SystemTray/Services/TerminalsManager.cs
MaxRis: Have added `GetLogFilePath` method with caching

MaxRis:
@felixse @mjs this, hopefully, is ready to be merged

FluentTerminal.SystemTray/Services/TerminalsManager.cs
```diff
@@ -8,36 +9,97 @@
...
+        private Dictionary<byte, string> _cachedLogPath = new Dictionary<byte, string>();
+
+        public TerminalsManager(ISettingsService settingsService)
+        {
+            _settingsService = settingsService;
+        }
+
         public void DisplayTerminalOutput(byte terminalId, byte[] output)
         {
+            if (_settingsService.GetApplicationSettings().EnableLogging && Directory.Exists(_settingsService.GetApplicationSettings().LogDirectoryPath))
```
felixse: Maybe we should cache this. Might cause up to two disk reads on every output

MaxRis: @felixse good catch! have rebased whole changelist, thanks

FluentTerminal.SystemTray/Services/TerminalsManager.cs
```diff
@@ -8,36 +9,98 @@
...
+        private Dictionary<byte, string> _cachedLogPath = new Dictionary<byte, string>();
+
+        public TerminalsManager(ISettingsService settingsService)
+        {
+            _settingsService = settingsService;
+        }
+
         public void DisplayTerminalOutput(byte terminalId, byte[] output)
         {
+            var appSettings = _settingsService.GetApplicationSettings();
```
felixse: still not happy with this. We should keep this as a field and update it on change (I think the UWP is already informing the SystemTray when ApplicationSettings have changed)

MaxRis: Unfortunately, there is no special Request class to inform SystemTray about changed settings or can't see any other notification mechanism. Going to add SettingsChangedRequest to be sent from UWP to Tray process.

MaxRis: @felixse here we go https://github.com/felixse/FluentTerminal/pull/438/commits/49b5d4ef97295282f89b98925779b09d1461b72e

MaxRis:
@felixse could you please check the the changes one more time, thanks!

PR #452, opened 2019-07-12
Run background process to keep conhost audio session active.

MaxRis:
"Mute Terminal" functionality enhancement discussed in previous PR https://github.com/felixse/FluentTerminal/pull/437#issuecomment-509979156

- Tray process on startup searches for available conhost audio sessions. If no audio session is available, detached cmd.exe process is spawned to play bell sound. Tray process searches for conhost audio session again and sets relevant mute state.
- Since first spawned cmd.exe process seats in OS background, Fluent will play only single bell sound on very first launch of Tray process (and will not play sound on "Mute Terminal" option change or on each run of Tray process as it does currently).

### DIFF
--- a/FluentTerminal.SystemTray/Utilities.cs
+++ b/FluentTerminal.SystemTray/Utilities.cs
@@ -569,48 +569,49 @@ namespace FluentTerminal.SystemTray
             }
         }
 
-        internal static void MuteTerminal(bool mute)
+        private static bool MuteConhost(bool mute)
         {
-            bool? deviceMuted = VolumeControl.GetDefaultAudioEndpointMute();
-            if (deviceMuted != true)
+            bool audioSessionFound = false;
+            foreach (Process conhost in Process.GetProcessesByName("conhost"))
             {
-                VolumeControl.SetDefaultAudioEndpointMute(true);
-            }
+                bool? isAudioSessionMuted = VolumeControl.GetAudioSessionMute(conhost.Id);
+                if (isAudioSessionMuted != null && isAudioSessionMuted != mute)
+                {
+                    VolumeControl.SetAudioSessionMute(conhost.Id, mute);
+                    Logger.Instance.Debug($"Mute state for conhost process with id={conhost.Id} was set to {mute}.");
+                }
 
+                if (isAudioSessionMuted != null)
+                {
+                    audioSessionFound = true;
+                }
+            }
+            return audioSessionFound;
+        }
+
+        private static void SpawnConhostProcess()
+        {
             Process cmdProcess = new Process();
             cmdProcess.StartInfo.FileName = "cmd.exe";
             cmdProcess.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
             cmdProcess.StartInfo.Arguments = "/k \x07";
             cmdProcess.Start();
 
-            bool wasConhostMuted = false;
-            int attempt = 0;
-            while (wasConhostMuted == false && attempt++ < 2 && !cmdProcess.HasExited)
-            {
-                cmdProcess.WaitForExit(1500);
+            cmdProcess.WaitForExit(1500);
+        }
 
-                foreach (Process conhost in Process.GetProcessesByName("conhost"))
-                {
-                    try
-                    {
-                        bool? isMuted = null;
-                        isMuted = VolumeControl.GetAudioSessionMute(conhost.Id);
-                        if (isMuted != null && isMuted != mute)
-                        {
-                            VolumeControl.SetAudioSessionMute(conhost.Id, mute);
-                            wasConhostMuted = true;
-                            Logger.Instance.Debug($"Mute state for conhost process with id={conhost.Id} was set to {mute}.");
-                        }
-                    }
-                    catch (Exception) { }
-                }
+        internal static void MuteTerminal(bool mute)
+        {
+            bool audioSessionFound = MuteConhost(mute);
+
+            if (audioSessionFound == true)
+            {
+                return;
             }
 
-            cmdProcess.Kill();
-            if (deviceMuted != true)
-            {
-                VolumeControl.SetDefaultAudioEndpointMute(false);
-            }
+            SpawnConhostProcess();
+
+            MuteConhost(mute);
         }
     }
 }


FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
...
-            cmdProcess.StartInfo.FileName = "cmd.exe";
-            cmdProcess.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
-            cmdProcess.StartInfo.Arguments = "/k \x07";
-            cmdProcess.Start();
-
-            bool wasConhostMuted = false;
+            bool audioSessionFound = false;
             int attempt = 0;
-            while (wasConhostMuted == false && attempt++ < 2 && !cmdProcess.HasExited)
+            while (audioSessionFound == false && attempt++ < attemptCount)
```
mjs:
```suggestion
            while (!audioSessionFound && attempt++ < attemptCount)
```

FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
...
                 foreach (Process conhost in Process.GetProcessesByName("conhost"))
                 {
                     try
                     {
-                        bool? isMuted = null;
-                        isMuted = VolumeControl.GetAudioSessionMute(conhost.Id);
-                        if (isMuted != null && isMuted != mute)
+                        bool? isAudioSessionMuted = null;
+                        isAudioSessionMuted = VolumeControl.GetAudioSessionMute(conhost.Id);
+                        if (isAudioSessionMuted != null && isAudioSessionMuted != mute)
```
mjs: There's no need for the null check. C# will do the right thing here if this is just:
```suggestion
                        if (isAudioSessionMuted != mute)
```

FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
...
             }
+            return audioSessionFound;
+        }
+
+        internal static void MuteTerminal(bool mute)
+        {
+            bool audioSessionFound = MuteConhost(mute, 1);
+
+            if (audioSessionFound == true)
+                return;
```
mjs: put braces around this as per the project's style

FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
...
-                            wasConhostMuted = true;
                             Logger.Instance.Debug($"Mute state for conhost process with id={conhost.Id} was set to {mute}.");
                         }
+
+                        if (isAudioSessionMuted != null)
+                        {
+                            audioSessionFound = true;
+                        }
                     }
                     catch (Exception) { }
```
mjs: Is hiding exceptions like this really a good idea? Can you be more specific about the kind of exceptions you're wanting to squash here? What about logging?

FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
...
+            cmdProcess.Start();
+
+            cmdProcess.WaitForExit(1500);
+
+            MuteConhost(mute, 2);
 
-            cmdProcess.Kill();
             if (deviceMuted != true)
             {
                 VolumeControl.SetDefaultAudioEndpointMute(false);
```
mjs: This should probably be a in try-finally to ensure the audio endpoint is always unmuted even if something goes wrong.

FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
...
+                VolumeControl.SetDefaultAudioEndpointMute(true);
+            }
+
+            Process cmdProcess = new Process();
+            cmdProcess.StartInfo.FileName = "cmd.exe";
+            cmdProcess.StartInfo.WindowStyle = ProcessWindowStyle.Hidden;
+            cmdProcess.StartInfo.Arguments = "/k \x07";
+            cmdProcess.Start();
+
+            cmdProcess.WaitForExit(1500);
```
mjs: How about extracting the code for playing a beep to it's own method? (readbility)

FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
             }
         }
 
-        internal static void MuteTerminal(bool mute)
+        private static bool MuteConhost(bool mute, int attemptCount)
```
mjs: Why are there different `attemptCount` values for different call sites?

FluentTerminal.SystemTray/Utilities.cs
MaxRis: `null` there means that audio session is not found for the process and we shouldn't call `SetAudioSessionMute` for it.

I haven't checked yet, but `if (isAudioSessionMuted != mute)` seems will result in `true` every time when `isAudioSessionMuted` is null and it's not what expected because of above.

FluentTerminal.SystemTray/Utilities.cs
```diff
@@ -569,44 +569,58 @@ internal static void SaveFile(string path, string content)
...
+        {
+            bool audioSessionFound = MuteConhost(mute, 1);
+
+            if (audioSessionFound == true)
+                return;
+
+            bool? deviceMuted = VolumeControl.GetDefaultAudioEndpointMute();
+            if (deviceMuted != true)
+            {
+                VolumeControl.SetDefaultAudioEndpointMute(true);
```
MaxRis: @mjs what do you think if to remove muting of audio device completely and let it play bell sound? So far it's 3-4 reports from different people to explain why the sound suddenly disappears.

FluentTerminal.SystemTray/Utilities.cs
mjs: Sorry, you're right. My suggestion would only work for `==`, not `!=`. Ignore this!

FluentTerminal.SystemTray/Utilities.cs
mjs: Yeah, I saw those reports. 

If the bell plays then people will probably complain about that... it might be less confusing than muting I guess.

[Terminus](https://www.termius.com/) lets the user disable the terminal bell (or use a visual bell) and they use xtermjs too. It might be good to dig into how they do it.

MaxRis: @mjs will check Terminus. Initially, I have seen that Fluent passes some option to xterm to configure bell, but I haven't seen any difference in real behavior.

MaxRis: https://github.com/felixse/FluentTerminal/blob/96ac5cd49fb1f4e5f6c79e7932fe4c029f0f15e8/FluentTerminal.Client/src/index.ts#L50

mjs: Never mind about Terminus. I've just played around with it and its terminal beep suppression doesn't work.

FluentTerminal.SystemTray/Utilities.cs
MaxRis: Done

FluentTerminal.SystemTray/Utilities.cs
MaxRis: Finally, I was not able to get any exception on 1809/1903 and just removed try/catch block

FluentTerminal.SystemTray/Utilities.cs
MaxRis: Have removed muting of device as it was discussed ( in next PR will attempt to mute audio session on its early start )

FluentTerminal.SystemTray/Utilities.cs
MaxRis: have removed looping completely

MaxRis:
@felixse please take a look to have this merged, thanks

PR #467, opened 2019-07-30
Web browser like tabs

MaxRis:
Related to https://github.com/jumptrading/FluentTerminal/issues/109

- TabBar.xaml extended to support tabs drag and drop events to allow tabs dropping to another window or to tear off tab and create new window.
- Pause and resume TerminalSession's std output reading while xterm.js is not listening.
- Enable AlwaysShowTabs Settings option by default.

Known issue to be addressed in next step:
Not whole state of xterm.js is restored on drop of tab for restored buffer content (but new received terminal session is displaying fine).

### DIFF
--- a/FluentTerminal.App/Views/MainPage.xaml.cs
+++ b/FluentTerminal.App/Views/MainPage.xaml.cs
@@ -98,14 +98,6 @@ namespace FluentTerminal.App.Views
             }
         }
 
-        private void TabView_TabDraggedOutside(object sender, Microsoft.Toolkit.Uwp.UI.Controls.TabDraggedOutsideEventArgs e)
-        {
-            if (e.Item is TerminalViewModel model)
-            {
-                ViewModel.TearOffTab(model);
-            }
-        }
-
         private async void TabView_Drop(object sender, DragEventArgs e)
         {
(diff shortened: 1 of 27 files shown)


FluentTerminal.App/Views/MainPage.xaml.cs
```diff
@@ -95,5 +97,44 @@ private void UpdateLayoutMetrics()
                 PropertyChanged(this, new PropertyChangedEventArgs(nameof(CoreTitleBarPadding)));
             }
         }
+
+        private void TabView_TabDraggedOutside(object sender, Microsoft.Toolkit.Uwp.UI.Controls.TabDraggedOutsideEventArgs e)
```
felixse: leftover of TabView?

felixse:
Awesome 👍

FluentTerminal.App/Views/MainPage.xaml.cs
MaxRis: yes, sorry, good catch!
TabView worked pretty well except that it doesn't provide a template to customize whole tab header area for specific state (hovered, pressed, etc) and option to display tabs headers at the bottom of page. To have that desired customization it seems it's needed to redefine too much functionality of TabView's high level templates.

PR #469, opened 2019-08-02
Add some Chinese translations😝

skyel1u:


felixse:
Awesome, thanks 👍 
You can also sign up here for a more comfortable way to help with the chinese translation (we still got some keys remaining): https://poeditor.com/join/project/TfKNIt826z

PR #477, opened 2019-08-10
FIXED: clicking on about in settings does not deselect previous item

### LINKED ISSUE #465: Opening about & clicking previous tab won't work
# How to reproduce?

1. open the settings window
2. click on general (if not already displayed)
3. click on the about icon
4. click on the general icon

# What should happen?
General section should be displayed.

# What happens?
Nothing: about section is still displayed.

**Note**: the bug is reproducible by replacing `general` with any other section.

This happens with version 0.5.3.0 on Windows 10 with May 2019 update.

warpdesign:
Fixes #465 I reported a few days ago.

Instead of dynamically adding a new item, adding it, setting and unsetting it again, I just add an hidden item that gets selected when about page is displayed.


felixse:
awesome 👍

PR #481, opened 2019-08-12
Saving command history only on actually successful session

peske:
This PR solves https://github.com/jumptrading/FluentTerminal/issues/171


peske:
I've also added here a fix for https://github.com/jumptrading/FluentTerminal/issues/173 bug.

PR #482, opened 2019-08-13
Ensure processes are killed for ConPTY sessions

mjs:
- Clarified Dispose logic for ConPTY Terminal and Process
- Disposing a ConPTY Process now kills it and its children
- ConPTY Terminal now cleans up everything in its Dispose
- Wired up missing Dispose calls in various finalizers
- Removed ConPTY CTRL_CLOSE_EVENT handling. This event never seems to actually happen and the Terminal should get cleaned up eventually anyway.

Fixes https://github.com/jumptrading/FluentTerminal/issues/94

### DIFF
--- a/FluentTerminal.SystemTray/FluentTerminal.SystemTray.csproj
+++ b/FluentTerminal.SystemTray/FluentTerminal.SystemTray.csproj
@@ -79,6 +79,7 @@
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Core" />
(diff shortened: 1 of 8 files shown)


mjs:
@felixse please don't merge this yet. On reflection, I'd like to think about some of these changes some more.

mjs:
@felixse This is good to go now. I wanted to review the Dispose logic for ConPTY related session some more and have made further improvements.

PR #484, opened 2019-08-13
Limit mute looping to 3 seconds timeout. More debug logs.

MaxRis:
Related to https://github.com/jumptrading/FluentTerminal/issues/180

@felixse it was discovered cruel CPU impact related to my recent changes for muting of audio sessions. This changelist removes "infinity" loops and adds back only 3 seconds wait timeout for such tasks as:

- Finding of spawned conhost.exe process by its parent process id
- Muting conhost's audio session (we able just loop the check for audio session availability)


MaxRis:
@felixse thank you for merging! 👍

PR #492, opened 2019-08-22
Multi-language support: Add fr-FR

FJduFou:
all translate in french, I do not know which modifier to change to take into account if it is not automatic


felixse:
Merci 👍 
We are managing our localization in [POEditor](https://poeditor.com/join/project/TfKNIt826z), please sign up if you want to help keep this up to date

PR #505, opened 2019-08-28
Crash on close; Resizing leads to freezed terminals.

MaxRis:
- Reset Action reference from DebouncedAction instance.
- Prevent crash on close of uninitilized terminal.

Related to https://github.com/jumptrading/FluentTerminal/issues/204, https://github.com/jumptrading/FluentTerminal/issues/194

https://github.com/jumptrading/FluentTerminal/issues/204 was introduced by recent changes to avoid memory leaks.

### DIFF
--- a/FluentTerminal.App.ViewModels/TerminalViewModel.cs
+++ b/FluentTerminal.App.ViewModels/TerminalViewModel.cs
@@ -118,7 +118,7 @@ namespace FluentTerminal.App.ViewModels
             TabThemes = new ObservableCollection<TabTheme>(SettingsService.GetTabThemes());
             TabTheme = TabThemes.FirstOrDefault(t => t.Id == ShellProfile.TabThemeId);
 
-            CloseCommand = new RelayCommand(CloseTab, CanExecuteCommand);
+            CloseCommand = new AsyncCommand(CloseTab, CanExecuteCommand);
             CloseLeftTabsCommand = new RelayCommand(CloseLeftTabs, CanExecuteCommand);
             CloseRightTabsCommand = new RelayCommand(CloseRightTabs, CanExecuteCommand);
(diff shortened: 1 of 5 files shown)


FluentTerminal.App.ViewModels/TerminalViewModel.cs
```diff
@@ -394,6 +400,21 @@ private void CloseSearchPanel()
...
+        {
+            return CanExecuteCommand();
+        }
+
+        private bool CanExecuteCommand()
+        {
+            return (Initialized == true && _disposalRequested == false);
+        }
+
+        private async void CloseTab()
```
felixse: this looks dangerous, I think we should use `AsyncCommand` like for `EditTitleCommand`

MaxRis: @felixse will it be fine just to revert back to anonymous definition on CloseCommand `async () => await TryClose().ConfigureAwait(false)` ? P.S. Could you please give more general details why it might be dangerous, thanks!

MaxRis: @felixse hmmm ... if I revert back to `CloseCommand = new RelayCommand(async () => await TryClose().ConfigureAwait(false), CanExecuteCommand);` I'm starting to receive again crashes from random places (if to close fast just opened tabs).

felixse: the anonymous action he had before was bad as well. Can you try with `AsyncCommand` and `async Task CloseTab()` signature?

MaxRis: Sure, going to give it a try

MaxRis: @felixse seems AsyncCommand works good, have pushed new commit with it, thanks

Issue #443, opened 2019-07-10
The layout in setting is overlapping


antoinebou13:
![image](https://user-images.githubusercontent.com/13888068/60988661-062f1b00-a312-11e9-8fb2-04e8ff071f68.png)

jenia90:
Hi, I would like to contribute to this project and I have a fix for this bug. This would be my first ever contribution, so I want to ask if there's anything special I should do when making pull request?

PR #517, opened 2019-09-08
Fixed overlap in Settings page.

jenia90:
This is a fix for [#443](https://github.com/felixse/FluentTerminal/issues/443).


felixse:
Thanks for looking into this. I think it might actually be enough to set the width of the first column to 160, so that it is in line with the grid defined in line 19. Mind if I push some changes here?

jenia90:
Oh yeah, I changed that by mistake when adding more columns in VS Blend. And I definitely don't mind.

PR #530, opened 2019-09-25
Fixed a typo in README.md

Neraeh:
Changes `Laguages` to `Languages` in README.md  
That's not a real issue but I think keeping the README clean is a nice thing to do for newcomers


felixse:
Good catch, thank you 👍

PR #542, opened 2019-10-07
Fixed a memory leak with opening / closing windows.

peske:
It fixes the part of https://github.com/jumptrading/FluentTerminal/issues/190 related to windows. It looks though that we still have a leak with opening / closing tabs.


felixse:
Nice 👍

Issue #539, opened 2019-10-03
Change the background image


ustcfdm:
For now the background is pure color. Can we set an image as the background? Here are some examples.

![image](https://user-images.githubusercontent.com/12667807/66150136-28b39600-e5da-11e9-87da-504994eed8e1.png)

![image](https://user-images.githubusercontent.com/12667807/66150291-74663f80-e5da-11e9-956e-b3cff3d77f5f.png)

YohskDista:
Hi,
I will be happy to work on this issue. Do you think this is part of the theme or it should be in the in the settings (section `Terminal`) ? I think it's more logical to put it in the `Terminal` section but I just want to know your opinion about it.

Also for the image selection, do we need to save it in a specific folder in case of the deletion of the image ?

Best regards.
Leonardo

felixse:
Thanks for picking this up 👍 

I think making it a part of theme makes it more flexible (like different themes with different images for different profiles). In regard of saving I would just put it somewhere in roaming storage. Also think about theme import/export. We could base64 encode it into the flutetheme file.

YohskDista:
No problem, it's an interesting project 👍 

Ok to put the images in the theme but I think it's not a good idea to put the images in the roaming storage. Microsoft says that the roaming storage it's for a little amount of data (like the user settings, or little files) and the images that the user can import could be very heavy.

Also I was thinking to this solution when we import an image :

1. In the theme settings we can browse to files and select a picture
1. Copy the selected image in the LocalCache folder so that we can show the image as preview in the settings window.
1. When the user click on `Save` button we will move the image in a `Image` folder located in the `Local` storage file.

Otherwise to avoid the step with the `LocalCache` folder we need to add capabilities to the application so that we can access to the user documents.

What do you think about this ?

felixse:
I guess images are still considered small enough. Problem is that themes are stored in roaming settings, so people might end up with missing background images on other computers.

I think we can skip the LocalCache step by just keeping it in memory until we save it.

YohskDista:
My bad, the images can be put in the roaming folder according to Microsoft doc :+1:

The problem of the preview image in the settings it's that we cannot show files that aren't in the folder of the application unless we activate the `broadFileSystemAccess` ([StackOverflow](https://stackoverflow.com/questions/54091707/how-to-display-image-form-local-forder-uwp)). That's because our URI is a `C:\....\file1.png` and it should be more like `ms-appdata:///.../file1.png`

[...]

PR #545, opened 2019-10-13
Add theme background image feature

YohskDista:
# Goal

The goal of this PR is to be able to select and show a background image for a custom theme and have a more fancy terminal. This refers to the issue #539.

# What was done

### Add / Edit the theme

Firstly we add a browse button when we create or edit a theme so that the user can select an image in a specific folder of his computer. Once the image is selected we will put it in the `LocalCache/BackgroundImageTmp` folder so that we can show a preview to the user.

We need to do this because our application cannot access or show an image that is outside its application folder unless we gave the application the specific capabilities.
To do a preview of the background image we must have a background color below the toolbar and the color configuration otherwise we didn't see anything because of the image.

When we save the theme we will remove all the image in the temporary folder and save the selected one in the `Roaming` folder.

We create an interface to manage the selection of these image functionnality called `IImageFileSystemService`. That allow us to separate the file system managment from the image theme managment.

### Remove the background image

We put a button to be able to remove the background image any time we want. If we click on the cancel button after we deleted the image don't worry, we don't loose your precious image :)

### Import / Export

We add the possibility to import and export the theme with the background image (if it exists). When we export it we encode the image in a base64 string and decode it when we import the theme.

To export/import the theme we need to create an `ExportedTerminalTheme` object that contains a `TerminalTheme` and the encoded image string. We did this because it was a pitty to change the `TerminalTheme` object for this encoded image that we need only for the import/export theme.

### Show the background image

Now we can show the background image of the selected theme directly in the `TerminalView`. Be carefull to select a good image so that you can still see the text of the console ;)


felixse:
Looks good so far, nice work 👍 

Some things I have noticed:
* After selecting an image and saving: If you select another theme and then go back to the just edited one, the background disappears
* The background colors for the the theme page look a bit odd. I don't think we need them. If you cannot read your text there anymore it's a good indicator that your image might not be a suitable background.
* The image is not keeping it's aspect ratio. Adding `Stretch.UniformToFill` to the Brush might help.

YohskDista:
Thanks ! I've made the suggestion that you were talking. I let you check if it's ok for you now 👍

Leonetienne:
This is such a cool feature! Thanks a lot! :)

PR #562, opened 2019-10-26
Default to WinPty settings option.

MaxRis:
Moves back global option to have preferred pty mode globally (ConPty or WinPty).

Related to https://github.com/jumptrading/FluentTerminal/issues/223

It allows to save few seconds of time on each opening of remote machine from SSH dialogs.


MaxRis:
@felixse could you please take a look at this PR. Having Settings option to switch globally between ConPty/WinPty is highly desired by those who still prefer WinPty over ConPty.

PR #567, opened 2019-11-02
Added hebrew translation

jenia90:
I've translate most of the strings on POEditor and imported the Resource.resw file into the project. 
However, the app might need some UI work for RTL languages such as Hebrew or Arabic.


felixse:
nice 👍

PR #580, opened 2019-11-16
Tab-selection in "Quick Launch" dialog

peske:
Solves https://github.com/jumptrading/FluentTerminal/issues/236

I've had to implement a hack to enforce Tab-selection.


peske:
I've just noticed a bug here. Closing this one, and will reopen soon.

PR #585, opened 2019-11-19
"Recent" menu implemented

peske:
Solves https://github.com/jumptrading/FluentTerminal/issues/235


peske:
image

peske:
I've also added one more term in POEditor: "Recent". You'll need translations...

PR #591, opened 2019-11-25
It's "brazilian", not "brasilian"

paulo-e:
If that's wrong in any other file I would ask to fix there too


felixse:
Should be the only file where this is wrong, thanks 👍

PR #594, opened 2019-11-25
Menu items now honor "Open new terminals as" settings

peske:
**Please merge https://github.com/felixse/FluentTerminal/pull/589 first!** This one builds upon it.

Solves https://github.com/jumptrading/FluentTerminal/issues/242. Menu items now open new tab or new window depending on "Open new terminals as" settings option value. The same stands for all menu items: root ones ("Default Profile", "Remote Connect..." and "Quick Launch...") as well as for items from "Recent >" submenu.

@felixse, I've had to change menu item text because "New tab" aren't applicable anymore - now they can also open new window. But changing the text of the items is now trivial - it's in the resources anyway, so once you have the final decision, you can change easily. Nevertheless, if nothing else, these ones are much better than the previous ones.

image


peske:
I've additionally replaced "Default Profile" text with the actual profile name:
image

As everything else there, it also updates dynamically.

@felixse, even if you're still for "New Window"/"New Tab" cascading menus, I urge you to merge this one. I promise I'll create another PR with cascading tab/window menus, so you'll be able to decide where to go. On top of this one, it'll take me less than hour to create the new cascading menu.

felixse:
brazil changes are mixed up, but I will export them from poeditor soon enough

PR #598, opened 2019-12-01
Dispatcher usage improvements

peske:
Part of https://github.com/jumptrading/FluentTerminal/issues/237

The key changes are:
- Now we can specify priority of the action that will be scheduled to dispatcher (i.e. `CoreDispatcherPriority.Low`);
- `enforceNewSchedule` now defaults to `false` (I've reviewed all the usages and concluded that it's safe);
- Now supports scheduling functions which are returning a value.


felixse:
So when exactly do we need to enforceNewSchedule?

PR #602, opened 2019-12-02
XtermTerminalView improvements and some async simplifications

peske:
Changes related to `XtermTerminalView` (https://github.com/jumptrading/FluentTerminal/issues/238):
- `_dispatcherJobs` blocking collection removed. Jobs are executed directly instead. No exceptions or other issues noticed in the new, simplified implementation.
- Handling resizing improved. Now it also covers some edge cases like resize during the initialization. Also output unblocking is better (more reliable) now - no scheduling with _hope_ that resize will finish. There's a bit more code than we had so far, but the workflow is much cleaner and obvious.

Changes related to https://github.com/jumptrading/FluentTerminal/issues/237:
- `async`/`await` simplified in some methods.

### DIFF
--- a/FluentTerminal.App.ViewModels/MainViewModel.cs
+++ b/FluentTerminal.App.ViewModels/MainViewModel.cs
@@ -145,7 +145,7 @@ namespace FluentTerminal.App.ViewModels
 
             if (changeMenu)
             {
-                ApplicationView.DispatchAsync(CreateMenuViewModel, CoreDispatcherPriority.Low, true);
+                ApplicationView.ExecuteOnUiThreadAsync(CreateMenuViewModel, CoreDispatcherPriority.Low, true);
             }
         }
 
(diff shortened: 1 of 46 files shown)


FluentTerminal.App.ViewModels/MainViewModel.cs
```diff
@@ -459,18 +482,14 @@ private void Terminal_CustomTitleChanged(object sender, string e)
 
         private void Terminal_ShellTitleChanged(object sender, string e)
```
felixse: remove stub or use this one instead of the more generic propertychanged again (the custom event is more readable in my opinion)

peske: Huh, I've just removed `CustomTitleChanged` event. Not sure why you think that the custom event is more reliable? `PropertyChanged` fires anyway, and covers both `ShellTitleChanged` and `CustomTitleChanged`.
I've removed the stub in one of my following PRs.

peske: No big difference (performance or anything) between `PropertyChanged` and custom events, but I believe that the code is simpler (two `event`s removed, subscribe/unsubscribe, different handlers...)

peske: You can see all the benefits in my following PR https://github.com/felixse/FluentTerminal/pull/604

PR #608, opened 2019-12-03
Overlay binding errors fixed

peske:
Fixes https://github.com/jumptrading/FluentTerminal/issues/241

I'm still confused why we were getting these errors, but the only way for me to fix them was to inject `OverlayViewModel` through constructor to `OverlayView`. This change slightly improves complexity, but I strongly believe that it's better than leaving binding errors there. In my knowledge from WPF binding errors can significantly affect UI performances.


peske:
Please don't merge just yet - I'll offer another alternative in minutes.

peske:
https://github.com/felixse/FluentTerminal/pull/609 is probably a better solution. Closing this one.

PR #618, opened 2019-12-04
Improving TabBar.xaml.cs

peske:

The new, significantly simplified version of `OnListViewSelectionChanged` has the same effect as the old version. Yet `Task.Run...` removed, and it's much simpler.


felixse:
I think the hack you removed stopped the tab bar scroll buttons from working

peske:
I'll check and let you know...

peske:
New PR is created: https://github.com/felixse/FluentTerminal/pull/668

Closing this one.

PR #619, opened 2019-12-04
Improving TerminalBridge

peske:

Will post another comment (see below) which describes what was wrong here.

### DIFF
--- a/FluentTerminal.RuntimeComponent/WebAllowedObjects/TerminalBridge.cs
+++ b/FluentTerminal.RuntimeComponent/WebAllowedObjects/TerminalBridge.cs
@@ -9,19 +9,15 @@ namespace FluentTerminal.RuntimeComponent.WebAllowedObjects
 {
     internal static class EventDispatcher
     {
-        private static CoreDispatcher _dispatcher => CoreWindow.GetForCurrentThread()?.Dispatcher;
-
         public static async void Dispatch(Action action)
         {
-            // already in UI thread:
-            if (_dispatcher == null || _dispatcher.HasThreadAccess)
+            if (CoreWindow.GetForCurrentThread()?.Dispatcher is { } dispatcher)
             {
-                await Task.Run(action);
+                await dispatcher.RunAsync(CoreDispatcherPriority.Normal, () => action());
             }
-            // not in UI thread, ensuring UI thread:
             else
             {
-                await _dispatcher.RunAsync(CoreDispatcherPriority.Normal, () => action());
+                await Task.Factory.StartNew(action);
             }
         }
     }


FluentTerminal.RuntimeComponent/WebAllowedObjects/TerminalBridge.cs
```diff
@@ -9,19 +9,15 @@ namespace FluentTerminal.RuntimeComponent.WebAllowedObjects
 {
     internal static class EventDispatcher
     {
-        private static CoreDispatcher _dispatcher => CoreWindow.GetForCurrentThread()?.Dispatcher;
-
         public static async void Dispatch(Action action)
         {
-            // already in UI thread:
-            if (_dispatcher == null || _dispatcher.HasThreadAccess)
+            if (CoreWindow.GetForCurrentThread()?.Dispatcher is { } dispatcher)
```
felixse: what is `{  }` ? Never seen this before

peske:
@felixse, there are several issues with the code I've changed. I'll describe these with my best intention to give you a chance to extend your knowledge. (Myself, I always love to learn something new...)

So here's the code in question (your original code):
```
internal static class EventDispatcher
{
    private static CoreDispatcher _dispatcher => CoreWindow.GetForCurrentThread()?.Dispatcher;

    public static async void Dispatch(Action action)
    {
        // already in UI thread:
        if (_dispatcher == null || _dispatcher.HasThreadAccess)
        {
            await Task.Run(action);
        }
        // not in UI thread, ensuring UI thread:
        else
        {
            await _dispatcher.RunAsync(CoreDispatcherPriority.Normal, () => action());
        }
    }
}
```
It has several issues, and at least one bug. Let's start from the beginning:

**1) `_dispatcher` property**

Since `_dispatcher` is a property (and it is thanks to `=>`), first it should be named as a property, capitalized (`Dispatcher`).

If `=>` is a typo, and you actually wanted it to be a field (with `=` instead of `=>`), then it's a bug. It would be a static field, meaning that it would have the same value throughout the process, meaning that the same `CoreDispatcher` would be used for all windows, which is obviously a bug. Another thing that's good to do is to add `readonly` for fields which doesn't change.

The next problem is the question should `_dispatcher` be a property in the first place. People are often confused with two implementation options which look the same at the first glance: property `Dispatcher` or method `GetDispatcher()` (both are returning the same thing). But there's one important difference:
- If you're returning always the same instance, probably created at the first call and then cached, for example, you should go with property implementation (`Dispatcher`);
- If you're always calculating / recreating the instance that will be returned, you should go with method implementation (`GetDispatcher()`).
This is because, if you implement it as a property, people will tend to use it as a property, meaning that they won't hesitate to use it multiple times within the same method, and even the same line. The way you're using it in your code is best example:
```
if (_dispatcher == null || _dispatcher.HasThreadAccess)
```
actually does the following:
```
if (CoreWindow.GetForCurrentThread()?.Dispatcher == null || CoreWindow.GetForCurrentThread().Dispatcher.HasThreadAccess)
```
which is obviously wrong. You're executing `CoreWindow.GetForCurrentThread()` twice in the same line, in total up to three times per method call.

But if you've implemented it as a method, then you (or some other user) would probably do it correctly, something like:
```
var dispatcher = GetDispatcher();
if (dispatcher == null || dispatcher.HasThreadAccess)
```
Long story in short: in this particular case you should go with method instead of property.

**2) The bug**

Take a look at your `if`/`else` implementation. What it would do if `_dispatcher.HasThreadAccess == true`? It would execute `Task.Run(action)`, which launches a new thread and executes `action` there. So effectively your `if`/`else` says: If we are in the UI thread, then execute the action in some other thread, but if we aren't in the UI thread, then schedule the action for execution in the UI thread, which doesn't make any sense, of course.

Again, pls sorry if you don't like to be criticized - I didn't want to be jerky...

FluentTerminal.RuntimeComponent/WebAllowedObjects/TerminalBridge.cs
peske: It has the same meaning as `is CoreDispatcher`, but ReSharper prefers shorter syntax `is { }`. It's basically checking if it is `null` or not.

peske:
Yep, one more thing: you should not use `Task.Run()`, but `Task.Factory.StartNew()` instead. Explanation: https://github.com/jumptrading/FluentTerminal/issues/237#issuecomment-557353593

PR #622, opened 2019-12-05
CS4014 warning solved

peske:

Changes are rather trivial. Mostly introducing discarding of the return value (`_ = ...`) in order to avoid warning (to annotate that we're intentionally using fire-and-forget pattern).


peske:
Hmmm... Not sure why it still contains `ITermThemeParser`. These changes are from the previous PR which is merged... Let me check that first.

peske:
It's OK now. Please merge.

PR #624, opened 2019-12-05
Reviewed async pattern throughout FluentTerminal.App project

peske:
**Merge https://github.com/felixse/FluentTerminal/pull/623 first!**

I've did a usual thing: removing `async`/`await` if they aren't needed, renaming methods by appending "Async", added `ConfigureAwait(false)` where applicable, removed `ConfigureAwait(true)` since it's assumed by default, and cleaning up some other things I encounter. Nothing significant changed (nothing that should affect the app features).

The only bit significant change is adding `StartupTaskStatus.EnabledByPolicy` value to the enumeration since it was missing there anyway.


peske:
@felixse , I've just resolved merging conflicts. You can merge...

PR #625, opened 2019-12-05
Async patterns in TrayProcessCommunicationService reviewed

peske:
**Merge https://github.com/felixse/FluentTerminal/pull/624 first!**

Again, I've did a similar thing: removing `async`/`await` where not needed, renaming async methods by adding "Async" suffix, and other small changes when I see something not-so-nice. **Nothing changed in the functionality.**

But this time I did something that you may don't like: I've replaced all `async`/`await` in `TrayProcessCommunicationService` with `Task.ContinueWith()` method. Some people don't like that, so I'll explain the difference.

Implementation with `async`/`await`:
```
public async Task<string> GetSshConfigDirAsync()
{
    var response = await GetSshConfigFolderAsync(false).ConfigureAwait(false);

    if (response.Success)
    {
        _sshConfigDir = response.Path;
    }

    return _sshConfigDir;
}
```
**Note that `ConfigureAwait(false)` is pretty important there** - otherwise we are wasting some resources.

Implementation with `ContinueWith()`:
```
public Task<string> GetSshConfigDirAsync()
{
    return GetSshConfigFolderAsync(false).ContinueWith(t =>
    {
        if (t.Result.Success)
        {
            _sshConfigDir = t.Result.Path;
        }

        return _sshConfigDir;
    }, TaskContinuationOptions.OnlyOnRanToCompletion);
}
```

Although the first variant has arguably cleaner code, I prefer the second because it's slightly more performant. But I have to admit that the difference in performances is very small (assuming that `ConfigureAwait(false)` is used). I've worked a lot on some trading software where microseconds count, so I always go for performances. But in case of desktop application, it really isn't important if something will take 5 or 10 milliseconds - users won't notice any difference.

In other words, if you prefer keeping `async`/`await` there - just let me know and I'll change quickly. And I'll know for future changes...


peske:
Replaced `ContinueWith()` implementations in `TrayProcessCommunicationService` with `async`/`await` implementations, as you requested.

peske:
@felixse Resolved merging conflicts here also. Please merge.

PR #635, opened 2019-12-26
Improved terminal output transfer

peske:
Solves https://github.com/jumptrading/FluentTerminal/issues/258, related to https://github.com/jumptrading/FluentTerminal/issues/250 and https://github.com/jumptrading/FluentTerminal/issues/257.

You'll probably want us to discuss before merging. But long story short: this speeds up output 6-7 times in case when there's a lot of output. See https://github.com/jumptrading/FluentTerminal/issues/258 for more details.


felixse:
I'm afraid this will not work. It might work on your machine because Visual Studio is enabling a local network loopback exception, but this will not run on a client device. (https://docs.microsoft.com/de-de/windows/iot-core/develop-your-app/loopback#loopback-and-the-debugger)

felixse:
I also don't think Protobuf is necessary here. JSON is only used for the non data messages, terminal input/output is already binary serialized.

But the the buffered reader sounds great 👍

peske:
BufferedReader will improve significantly even without the rest (protobuf /
ZeroMQ)

Data is binary serialized? Now I'm surprized! How it's transferred then?
We'll talk on Slack these days.

felixse:
https://github.com/felixse/FluentTerminal/blob/7251beb2f105b920af9c5dc2a68783d16b4a56d7/FluentTerminal.App.Services/Implementation/TrayProcessCommunicationService.cs#L228

Data is handled different from the other messages. No message object that gets serialized to JSON but instead we write the data directly to the ValueSet used by the AppServiceConnection. Which is then afaik binary serialized by the framework.

Communication will be a but limited since I'm attending 36C3 but I'll do my best to follow this

Issue #540, opened 2019-10-03
Can not define Environment Variables names in uppercase


jmparis:
The new feature added in the profiles: Environment variables has a flaw.
It's not possible to define a variable in uppercase.

Example:
USER     john
PATH    /c/msys64/usr/bin

both definitions are saved, and then pass to the shell as
user   john
path   /c/msys64/usr/bin

felixse:
Might be caused/fixed by: https://stackoverflow.com/questions/24143149/keep-casing-when-serializing-dictionaries

PR #644, opened 2020-01-04
Bug with env var names becoming lowercase fixed

peske:
Solves https://github.com/jumptrading/FluentTerminal/issues/261

Surprisingly for me, it turned out that `JsonConvert.SerializeObject`, with default settings **changes dictionary keys to lowercase!** I cannot believe that NewtonSoft JSON did such dumb mistake!


felixse:
Fixes #540

PR #659, opened 2020-01-31
adjustments that allow CI build step to run locally

marcpems:
some simple changes that seem to resolve the CI locally on my dev box.


felixse:
Do you know why it failed for x86 but worked with x64?

marcpems:
Merging threads…


Hmm, this has me intrigued. I can get the package to build with various changes, but I’m struggling to identify exactly what of the changes causes the build to work.

The command line I’m using:
msbuild -t:build -p:Configuration=Debug -p:Platform=x64 -restore -m

and removed the X86 target from the packaging project, as well as add/remove some redundant dependencies (add/remove the packaging project did this).
https://github.com/felixse/FluentTerminal/pull/659

CI works with this PR. I think it’s the X86 target that was causing the issue, but I’ve tried adding this back / removing etc and I just cant get a consistent response.
Anyway, the PR is there for you if you want to take it.

Marcus

felixse:
That's great, thanks 👍 

Did this also fix the packaging issues or is this another issue?

marcpems:
I’m not sure what the packaging issue was, can you be more specific?
With these changes I can generate packages through MSbuild (the CI command) and sideload them. I can also generate packages through VS and install them.

felixse:
But only for for x64?
I had issues similar to the CI when I tried to build packages locally. Maybe x86 was also causing this. I will try it and report back.

Is there anything we can do to continue offering x86 releases?

marcpems:
I’ve not looked at what is blocking X86.
I’ll have a look at that next week if you can wait.

felixse:
Yes sure, no need to hurry.

I just cloned a new repo with the latest changes and tried to built a package.
I can start a debug build, but when I try to create a package with VS I only see the configurations Store (AnyCPU) and Store (ARM64) for x64 architecture. Both obviously don't succeed. Did you test packaging with msbuild.exe directly?

PR #668, opened 2020-02-23
Improved TabBar.

peske:
Improved version of https://github.com/felixse/FluentTerminal/pull/618

Tested and works.


felixse:
Looks much better than my original hack 👍

PR #673, opened 2020-02-29
Disable WebView zooming to prevent crash on too small font size.

MaxRis:
Fixes https://github.com/jumptrading/FluentTerminal/issues/265

### DIFF
--- a/FluentTerminal.Client/src/index.ts
+++ b/FluentTerminal.Client/src/index.ts
@@ -175,6 +175,21 @@ window.createTerminal = (options, theme, keyBindings) => {
 
   window.onmouseup = (e) => defaultOnMouseUpHandler(e);
 
+  window.onkeydown = (e) => {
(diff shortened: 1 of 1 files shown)


FluentTerminal.Client/src/index.ts
```diff
@@ -175,6 +175,21 @@ window.createTerminal = (options, theme, keyBindings) => {
 
   window.onmouseup = (e) => defaultOnMouseUpHandler(e);
 
+  window.onkeydown = (e) => {
+    // Disable WebView zooming to prevent crash on too small font size
+    if ((e.ctrlKey && e.shiftKey && (e.keyCode === 189 || e.keyCode === 187)) ||
```
felixse: zoom also happens without shift

MaxRis: good catch! Have updated PR

Issue #669, opened 2020-02-25
BufferedReader causes issues with ConPty and PowerShell


felixse:
![8d2xmHQPBl](https://user-images.githubusercontent.com/30559812/75281972-85754e80-5810-11ea-9293-85e7e8b72592.gif)

Experiencing this with this exact combination. If I switch to WinPty or revert changes in ConPtySession that introduced the BufferedReader its behaving normally.

Output log:
[20200225084737086_Powershell.log](https://github.com/felixse/FluentTerminal/files/4251861/20200225084737086_Powershell.log)

vasiby:
Started to see something similar since xterm 4.2.0, even if I build 0.6.1 tag with the latest xterm, have to build it using npm ci, because npm install will download the latest xterm and cause this buffer issue

felixse:
not sure if this is the same problem. Do you experience this with ConPty or WinPty?

vasiby:
ConPTY and oh-my-posh module in powershell, only 0.6.1 branch is  stable with xterm 4.0.1.

peske:
I cannot reproduce this one, but I'm using standard prompt (`PS C:\Users\fatdragon>`). Can you check with the regular prompt? Can you send me your prompt function so that I can check?

felixse:
Same issues without my profile. What version of powershell are you testing this with? I'm on 7.0

peske:
Oh, I have it. It happens only for ConPTY, right?

PR #680, opened 2020-03-22
"Use buffer" profile settings implemented

peske:
Solves https://github.com/felixse/FluentTerminal/issues/669

### DIFF
--- a/FluentTerminal.SystemTray/Services/WinPty/WinPtySession.cs
+++ b/FluentTerminal.SystemTray/Services/WinPty/WinPtySession.cs
@@ -18,9 +18,12 @@ namespace FluentTerminal.SystemTray.Services.WinPty
         private TerminalsManager _terminalsManager;
         private Process _shellProcess;
         private BufferedReader _reader;
+        private bool _enableBuffer;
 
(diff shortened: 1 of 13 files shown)


FluentTerminal.SystemTray/Services/WinPty/WinPtySession.cs
```diff
@@ -18,9 +18,12 @@ public class WinPtySession : ITerminalSession
         private TerminalsManager _terminalsManager;
         private Process _shellProcess;
         private BufferedReader _reader;
+        private bool _enableBuffer;
```
felixse: I would rather scope this to conpty only, since we had no issues with wintpy and the buffer

peske: Since we are adding this option to profile settings, I believe that it should be independent on ConPTY / WinPTY. It's true that currently we're experiencing issues only with ConPTY, but it may happen that similar happens with WinPTY and particular tool.
Although I can disable this option for WinPTY, I believe that it would only introduce some confusion.

PR #682, opened 2020-03-24
Reconnect failed tab.

maxhora:
Fixes https://github.com/jumptrading/FluentTerminal/issues/222

Overlay to reconnect/close tab; Shortcut to reconnect tab.

### DIFF
--- a/FluentTerminal.App/Views/TerminalView.xaml
+++ b/FluentTerminal.App/Views/TerminalView.xaml
@@ -18,111 +18,149 @@
 
         <views:OverlayView x:Name="MessageOverlay" />
 
-        <controls:DropShadowPanel
-            Height="64"
-            Margin="12"
-            HorizontalAlignment="Right"
-            VerticalAlignment="Top"
-            ShadowOpacity="0.5"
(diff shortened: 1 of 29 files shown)


felixse:
I'm not a fan of the UI part of this. The tab item is not very easy to discover, plus this adds an almost always disabled item to the menu (which is also quite redundant for non ssh tabs).

How about some overlay similar to search/resize with two buttons (reconnect, close)?

MaxRis:
@felixse sounds reasonable, will try to update this PR with new overlay changes

FluentTerminal.App/Views/TerminalView.xaml
```diff
@@ -18,111 +18,149 @@
...
+                        ToolTipService.ToolTip="Close search panel">
+                        <Button.Content>
+                            <Viewbox Width="16" Height="16">
+                                <SymbolIcon Symbol="Cancel" />
+                            </Viewbox>
+                        </Button.Content>
+                    </Button>
+                </StackPanel>
+            </controls:DropShadowPanel>
+            <controls:DropShadowPanel
```
maxhora: @felixse  New `DropShadowPanel` is wrapped into StackPanel along with overlay for Search.

maxhora:
@felixse replaced new context menus with two buttons overlay to reconnect or close failed tab. Please let me know if any further adjustments make sense, thanks!

PR #707, opened 2020-04-08
Update README.md for Scoop

Restia666Ashdoll:
Scoop enables app sideloading from the manifest itself. Also, the manifest has now been updated for 0.7.0.0.


felixse:
Thanks 👍

TPXP:
Does that mean we should/can remove this line from Chocolatey's install script ?
https://github.com/pgalbraith/chocolatey-packages/blob/master/manual/fluent-terminal/tools/chocolateyinstall.ps1#L45

Issue #720, opened 2020-04-14
[Suggestion] Keyboard shortcut to zoom


Loreno10:
It would be great to have an option to change the font size with a keyboard shortcut (like `CTRL +` or `CTRL -`)

Loreno10:
@felixse I'd be happy to try to implement it, if it's not too difficult (I'm not too familiar with UWP and TS and I don't understand the magic that binds them together). If you could give me some hints what files I should modify, I will let you know if I can do it.

As I understand, what is needed is just a keyboard shortcut to increase the font size.

felixse:
I don't think this one is exactly beginner friendly, but you can have a try 😄 

My thoughts on this:
* add a new field to TerminalViewModel to track the current font size (initialize with the value form TerminalOptions)
* add 2 new Commands for increase and decrease and add some new defaults to DefaultValueProvider
* Extend TerminalViewModel.Terminal_KeyboardCommandReceived to handle the new commands (also validate that it does not go out of range)
* The nasty part: add a new method in index.ts that takes a new fontSize and pass it to term.setOption
* Extend XtermTerminalView with a call to this method
* Bonus: Use TerminalViewModel.Overlay to show the new font size

Loreno10:
@felixse Thanks for the recipe! :) I'll give it a try in the weekend I think.

zsbolyoczki:
Good idea, it would be also nice to zoom with CTRL + mouse wheel just like in mintty/wsltty.

Loreno10:
@felixse I added a draft PR https://github.com/felixse/FluentTerminal/pull/728
Unfortunately I wasn't able to finish this, I do not know how to extend `XtermTerminalView`. Project is quite big and I do not really understand the architecture :/
BTW, it would be awesome if you shared some more information about this project (architecture, technnologies, etc.).

@zsbolyoczki Yeah, that would be useful.

Making the project run was a big problem. I actually had to switch to another laptop. FIrst one just wasn't able to build the project for some reason. On the other laptop I went further, but still restore was not possible without `nuget.exe restore` (Visual Studio alone wasn't able to restore). Took me some hours to figure this out :O

PR #728, opened 2020-04-17
ADD base for increasing and decreasing font size with keys

Loreno10:
PR not finished!

I followed the steps from https://github.com/felixse/FluentTerminal/issues/720

THe thing I struggle with is "Extend XtermTerminalView with a call to this method". I do not know how to achieve this. In what way should I extend it?

Also, I'm not sure about "add some new defaults to DefaultValueProvider". `DefaultValueProvider` already contains method to get `TerminalOptions` - this object already contains `FontSize`. Is there a need to add anything to `DefaultValueProvider` then?


FluentTerminal.App.ViewModels/TerminalViewModel.cs
```diff
@@ -157,6 +160,10 @@ public void Restore(string data)
             PasteCommand = new AsyncCommand(Paste);
             CopyLinkCommand = new AsyncCommand(() => CopyTextAsync(HoveredUri), () => !string.IsNullOrWhiteSpace(HoveredUri));
             ShowSearchPanelCommand = new RelayCommand(() => ShowSearchPanel = true, () => !ShowSearchPanel);
+            IncreaseFontSizeCommand = new AsyncCommand(IncreaseFontSize, CanExecuteCommand);
+            DecreaseFontSizeCommand = new AsyncCommand(DecreaseFontSize, () => CanExecuteCommand() && FontSize > 1);
+
+            FontSize = _terminalOptions.FontSize;
```
Loreno10, line 166: Not sure about this one

FluentTerminal.App.ViewModels/TerminalViewModel.cs
```diff
@@ -264,6 +271,11 @@ public TabThemeViewModel BackgroundTabTheme
 
         public ICommand ShowSearchPanelCommand { get; private set; }
 
+        public ICommand IncreaseFontSizeCommand { get; private set; }
+
+        public ICommand DecreaseFontSizeCommand { get; private set; }
```
Loreno10, line 276: Where do we actually bind these commands to something?

felixse: We don't need commands in this case since the actions will be handled by xterm.js itself

felixse:
> PR not finished!
> 
> I followed the steps from #720
> 
> THe thing I struggle with is "Extend XtermTerminalView with a call to this method". I do not know how to achieve this. In what way should I extend it?

Something like `ChangeThemeAsync`. A method to call the method in the webview
 
> Also, I'm not sure about "add some new defaults to DefaultValueProvider". `DefaultValueProvider` already contains method to get `TerminalOptions` - this object already contains `FontSize`. Is there a need to add anything to `DefaultValueProvider` then?

You already added the new keybindings there. That's what this was about 😃

PR #812, opened 2020-07-13
Fix for lack of focus on search textbox.

RhavoX:
Focus was set before the control was visible most of the time. This fix will wait for the control to load if it is not visible. If it managed to load fast enough it will grab focus immediately instead.


felixse:
Looks great, thanks 👍

PR #814, opened 2020-07-14
fixes #761

### LINKED ISSUE #761: Duplicate tab should inherit the color
Currently if I have an open tab with a color and I duplicate it, the new tab starts without any color.

My workflow usually is to use colors to indicate which tabs are open in a particular folder, so I may have a group of tabs open on folder `/foo` that are blue and another group that are on folder `/bar` that are green. This would be very useful when opening a new tab.

Not sure if this should be default or just a config options

Thanks for this terminal, it's really great!

RhavoX:
Also copies title to other tabs


RhavoX:
Maybe some more description (sorry for being vague). This fix allows colors to be duplicated as well as tab title when duplication feature is used.

felixse:
No problem, just wanted to have a closer look at this one. Looks good 👍

PR #820, opened 2020-07-20
Fixed almost all warnings in the project. 

RhavoX:
Only language warning remain for the bundle project due to weird AZ language detection issues.


felixse:
Thanks for the cleanup 👍

PR #882, opened 2021-01-11
Fix explorer context integration directory param.

jareqpl:
I've found this bug while wondering why my terminal always opens in user directory. This change fixes it.


hanskokx (approved):

felixse:
This actually breaks it for me (at least when right clicking the explorer background, right clicking a directory is still working).

felixse:
Does the change in #884 solve your issue?

Issue #867, opened 2020-11-10
Fluent Terminal 0.7.3.0 x64 Windows 10 19041.610, wrong shell extension path open fluent terminal here


reidsneo:
Dear community,

I'm really love this terminal but I having issue with [terminal Integration](https://github.com/felixse/FluentTerminal/tree/master/Explorer%20Context%20Menu%20Integration) which is not working as expected
1. Right click > Open Fluent Terminal here
2. Displayed path is still on user path instead the current folder path where I right click the Fluent Terminal

I also tried to uninstall and reinstall the shell integration but the issue still there
Os : Windows 10 Home Single Language
Version :2004
Build : 19041.610

Thanks!

reidsneo:
The terminal integration is now works with [this patch](https://github.com/felixse/FluentTerminal/commit/6a123ed953242860c30c7f1279df29ad374a46e7) but the icon somewhat missing

Issue #719, opened 2020-04-14
"Open Fluent Terminal here" - no icon


Loreno10:
It would be cool to have an icon, like Hyper has

![image](https://user-images.githubusercontent.com/10273406/79214347-f9b09380-7e4a-11ea-9985-c235df5b484d.png)

PR #884, opened 2021-01-14
Made changes to install.bat and added install_with_icon.bat to fix issues #867 and #719 respectively

akanealw:
Updated install.bat to open in current directory.
Updated install.bat to have context menu entry when right-clicking on drives.
Added install_with_icon.bat with same changes as listed previously as well as adding an icon to the context menu items.
Uploaded icon file converted from official png logo for the install_with_icon.bat.
Added README for the install_with_icon.bat script.


felixse:
I think we should put the icon into Assets directory and then reference it at `C:\Program Files\WindowsApps\53621FSApps.FluentTerminal_0.7.4.0_x64__87x1pks76srcp\Assets` (there might by a variable for `C:\Program Files\WindowsApp`) to avoid files that will not be removed when uninstalling

akanealw:
Do you mean via my script or including the icon in the install package? I know that the WindowsApp directory is normally not accessible via normal administrator rights when running my script.

felixse:
Yes, you're right. I added a line in uninstall to remove the icon from `/WindowsApps` instead, should be equally effective

 Thanks for your work on this 👍

felixse (approved):

Issue #957, opened 2021-09-01
Crash when opening the settings...


BruceGeng2012:
Hi there,

When I try to open settings, still crash. Don't know why, can u plz help. 1st time to use the app. Love it...

OS Name:                   Microsoft Windows 10 Professional
OS Version:                10.0.19042 N/A Build 19042
System Type:               x64-based PC
Total Physical Memory:     16,351 MB
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 42 Stepping 7 GenuineIntel ~3301 Mhz

Thanks

beyond-danube:
OK, so issue is more unpleasant then it looks at first glance.

Long story short: there is no easy workaround.

Background, I have 2 laptops. 100% reproducible on one, and not reproducible on another. Exception happens in DWrite.dll, which is written in C++, an external unmanaged  library from CLR perspective. Exception is an Access Violation, so it cannot be catched or processed as a generic unhandled exception from UWP app and .NET above 4.0.

Exceptions happens on getting Fonts installed in system, so it would be OK just not to show fonts to a user in that case, but it's impossible due to case above. Logged issue Win2D, however I don't think it's going to be fixed any time soon https://github.com/microsoft/Win2D/issues/863

References on MSDN:
[HandleProcessCorruptedStateExceptionsAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.handleprocesscorruptedstateexceptionsattribute?redirectedfrom=MSDN&view=net-6.0)
[legacyCorruptedStateExceptionsPolicy](https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/file-schema/runtime/legacycorruptedstateexceptionspolicy-element?redirectedfrom=MSDN)

I tried to move service to FluentTerminal.App.Services.Implementation which is in .NET Standard 2.0, did not help obviously since method call happens from UWP anyway, it fails just in another place.

Other way to keep same functionality would be to use System.Drawing.Common with good old GDI+ to get installed fonts and [LOGFONT Structure](https://www.pinvoke.net/default.aspx/Structures/LOGFONT.html) hook to detect if font is monospaced or not. This will not work for similar reason, no platform-dependent libraries will work with UWP.

[...]

PR #993, opened 2022-01-21
Fixing #957

beyond-danube:
Details in comment https://github.com/felixse/FluentTerminal/issues/957#issuecomment-1018042944


felixse:
Thanks for the PR 👍 
Hopefully Microsoft will fix this in the future so that monspaced highlighting can return

felixse (approved):

